PR #17, opened 2021-03-18
Fix mdox-gen-exec issues with --help commands

### LINKED ISSUE #10: `mdox-gen-exec` Make it easier to use with --help
Usually  --help returns non 0 exit code. This fails to format.

Workaround like ````bash mdox-gen-exec="sh -c 'bingo --help || exit 0'"` works but it's ugly.

Wonder if we can improve the experience somehow.

saswatamcode:
This PR fixes the issue where commands with non 0 exit codes (usually `--help` commands) fail to format. Fixes #10 
Before this fix, the markdown below results in an error,
```
```bash mdox-gen-exec="go --help"
.
.
.
error:  first formatting phase for /Users/saswatamukherjee/web/mdox/test.md: run go --help: exit status 2
``` 
A workaround is using something like `mdox-gen-exec="sh -c 'go --help || exit 0'"` but this is not preferable. 
With this fix, a flag can be passed to allow non zero exit codes to format successfully, like below, 
```
```bash mdox-nonzero-exit-code mdox-gen-exec="go --help"
Go is a tool for managing Go source code.

Usage:

	go <command> [arguments]
...
```
This would also be suitable in case an error code block example needs to be generated.

### DIFF
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -19,6 +19,7 @@ const (
 	infoStringKeyLang = "mdox-gen-lang"
 	infoStringKeyType = "mdox-gen-type"
 	infoStringKeyExec = "mdox-gen-exec"
+	skipExitCodeCheck = "mdox-nonzero-exit-code"
 )
 
 type genCodeBlockTransformer struct{}
@@ -37,6 +38,7 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 		return nil, errors.Wrapf(err, "parsing info string %v", string(infoString))
 	}
 	infoStringAttr := map[string]string{}
+	exitCodeSkip := false
 	for i, field := range infoFiels {
 		if val := strings.Split(field, "="); val[0] == infoStringKeyLang || val[0] == infoStringKeyType || val[0] == infoStringKeyExec {
 			if i == 0 {
@@ -46,6 +48,10 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 				return nil, errors.Errorf("got %q without variable. Expected format is e.g ```yaml %q=<value> %q=<value2> . Got info string %q", val[0], infoStringKeyLang, infoStringKeyType, string(infoString))
 			}
 			infoStringAttr[val[0]] = val[1]
+		}
+
+		if field == skipExitCodeCheck {
+			exitCodeSkip = true
 		}
 	}
 
@@ -69,6 +75,9 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 		cmd.Stderr = &b
 		cmd.Stdout = &b
 		if err := cmd.Run(); err != nil {
+			if _, ok := err.(*exec.ExitError); ok && exitCodeSkip {
+				return b.Bytes(), nil
+			}
 			return nil, errors.Wrapf(err, "run %v", execCmd)
 		}
 		return b.Bytes(), nil
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/bwplotka/mdox/pkg/mdformatter"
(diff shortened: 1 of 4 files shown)


pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -19,6 +19,7 @@ const (
 	infoStringKeyLang = "mdox-gen-lang"
 	infoStringKeyType = "mdox-gen-type"
 	infoStringKeyExec = "mdox-gen-exec"
+	skipExitCodeCheck = "mdox-nonzero-exit-code"
```
bwplotka, line 22: I wonder... should we just don't care about exit codes? 🤔

bwplotka: Also variable name is not telling the same as string

bwplotka (commented):
Epic work, thanks!

Wonder... what would be the best experience here. Some suggestions:

1. What about opposite: have option `expect-exit-code` = `2` and by default don't check exit code?
2. Can we run quick test to ensure this? 🤗

pkg/mdformatter/mdgen/mdgen.go
saswatamcode: Initially, I did think that not checking for exit codes would be better, but then user wouldn’t be able to tell which command in a md code block had an actual error and that might get formatted without the user noticing. For eg `npm —versionn` is a typo which would come up as an error if we check the exit code. However, if we skip that then something entirely different than the version number might get formatted.
Would the better option be to have an `expect-exit-code` flag which will be matched if non 0 exit code occurs?

bwplotka: I think tag makes sense more like you did.

So what about by default expecting status code 0, but allow `expect-exit-code=X` to change that? 🤗

saswatamcode: Yup! That would be best. Implementing that along with a quick test! 😊

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -16,9 +17,10 @@ import (
 )
 
 const (
-	infoStringKeyLang = "mdox-gen-lang"
-	infoStringKeyType = "mdox-gen-type"
-	infoStringKeyExec = "mdox-gen-exec"
+	infoStringKeyLang     = "mdox-gen-lang"
+	infoStringKeyType     = "mdox-gen-type"
+	infoStringKeyExec     = "mdox-gen-exec"
+	infoStringKeyExitCode = "expect-exit-code"
```
bwplotka, line 23:
```suggestion
	infoStringKeyExitCode = "mdox-expect-exit-code"
```

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -38,12 +40,12 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 	}
 	infoStringAttr := map[string]string{}
 	for i, field := range infoFiels {
-		if val := strings.Split(field, "="); val[0] == infoStringKeyLang || val[0] == infoStringKeyType || val[0] == infoStringKeyExec {
+		if val := strings.Split(field, "="); val[0] == infoStringKeyLang || val[0] == infoStringKeyType || val[0] == infoStringKeyExec || val[0] == infoStringKeyExitCode {
```
bwplotka, line 43: Time for some `switch` here (:

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -38,12 +40,12 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
...
 			if i == 0 {
 				return nil, errors.Errorf("missing language info in fenced code block. Got info string %q", string(infoString))
 			}
 			if len(val) != 2 {
-				return nil, errors.Errorf("got %q without variable. Expected format is e.g ```yaml %q=<value> %q=<value2> . Got info string %q", val[0], infoStringKeyLang, infoStringKeyType, string(infoString))
+				return nil, errors.Errorf("got %q without variable. Expected format is e.g ```yaml %q=<value> %q=<value2> %q=<value2>. Got info string %q", val[0], infoStringKeyLang, infoStringKeyExitCode, infoStringKeyType, string(infoString))
```
bwplotka, line 48: same here

bwplotka (approved):
Amazing, looks good. Just some suggestions still.

bwplotka (commented):
Nice, thanks!

LGTM 💪🏽

Issue #18, opened 2021-03-27
Add `mdox-gen-file` to allow fetching from file


saswatamcode:
Add `mdox-gen-file` to allow for fetching code blocks from files with customizable options.

bwplotka:
As agreed in PR, let's show how sed can be leveraged (:

PR #19, opened 2021-03-27
Add mdox-gen-file

saswatamcode:
This PR adds an `mdox-gen-file` option which allows fetching code from local files into documentation like below. Addresses #18 
```
```go mdox-gen-file="main.go" mdox-gen-lines=47-56

```
This results in the following when formatted,
```
```go mdox-gen-file="main.go" mdox-gen-lines=47-56
	switch logFormat {
	case logFormatJson:
		return level.NewFilter(log.NewJSONLogger(log.NewSyncWriter(os.Stderr)), lvl)
	case logFormatLogfmt:
		return level.NewFilter(log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)), lvl)
	case logFormatCLILog:
		fallthrough
	default:
		return level.NewFilter(clilog.New(log.NewSyncWriter(os.Stderr)), lvl)
	}
. . .
```
If no line numbers are given like below, entire file is brought in by default (useful for blogs),
```
```go mdox-gen-file="main.go" mdox-gen-lines=47-56
```

### DIFF
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -159,7 +159,7 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 			line++
 		}
 		if err := scanner.Err(); err != nil {
-			return nil, errors.Errorf("scanner couldn't read: %v", err)
+			return nil, errors.Wrap(err, "scanner read")
 		}
 		return text, nil
 	}
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -7,6 +7,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"math"
 	"os"
 	"os/exec"
 	"strconv"
@@ -72,6 +73,12 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 		return code, nil
 	}
 
+	_, execkeyOk := infoStringAttr[infoStringKeyExec]
+	_, filekeyOk := infoStringAttr[infoStringKeyFile]
+	if execkeyOk && filekeyOk {
+		return nil, errors.Errorf("got ambiguous instructions. %q and %q can't be used together. Separate code blocks expected. Got info string %q", infoStringKeyExec, infoStringKeyFile, string(infoString))
+	}
+
 	// Code fence with command.
 	if execCmd, ok := infoStringAttr[infoStringKeyExec]; ok {
 		if len(infoStringAttr) > 2 {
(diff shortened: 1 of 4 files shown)


pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -21,6 +23,8 @@ const (
 	infoStringKeyType     = "mdox-gen-type"
 	infoStringKeyExec     = "mdox-gen-exec"
 	infoStringKeyExitCode = "mdox-expect-exit-code"
+	infoStringKeyFile     = "mdox-gen-file"
```
bwplotka, line 26: Hm, the only problem with metod is that what happens if one would specify both `mdox-gen-exec` and `mdox-gen-file`? (:

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -100,6 +111,53 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
...
+
+		scanner := bufio.NewScanner(file)
+		scanner.Split(bufio.ScanLines)
+		var text []byte
+
+		switch lineOk {
+		case true:
+			val := strings.Split(linesToCopy, "-")
+			start, errStart := strconv.Atoi(val[0])
+			if errStart != nil {
```
bwplotka, line 133:
```suggestion
			if err != nil {
```
Usually it's fine to shadow err variable if in short scope.

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -100,6 +111,53 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
...
+		}
+		defer file.Close()
+
+		scanner := bufio.NewScanner(file)
+		scanner.Split(bufio.ScanLines)
+		var text []byte
+
+		switch lineOk {
+		case true:
+			val := strings.Split(linesToCopy, "-")
```
bwplotka, line 131: I would propose the same semantics as Go slice, WDYT?
```suggestion
			val := strings.Split(linesToCopy, ":")
```

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -100,6 +111,53 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
...
+			}
+			line := 0
+
+			for scanner.Scan() {
+				if line >= start && line <= end {
+					text = append(text, scanner.Bytes()...)
+					text = append(text, "\n"...)
+				}
+				line++
+			}
```
bwplotka, line 151: Let's check `scanner.Err()` after this loop (same below)

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -100,6 +111,53 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
...
+		case false:
```
bwplotka, line 152: Can we reuse the above, but just use start 0 and end = `math.MaxInt64`? (:

bwplotka (commented):
Amazing! Some suggestions, plus some test needed, otherwise LGTM!

pkg/mdformatter/mdgen/mdgen.go
saswatamcode: I think an error needs to be shown in that case, so that one of them can be removed and there's no confusion later on what would get generated in that block. WDYT? (:

pkg/mdformatter/mdgen/mdgen.go
saswatamcode: Sure! I think this would be better!

pkg/mdformatter/mdgen/mdgen.go
saswatamcode: Sure. Will make the necessary changes!

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -100,6 +118,52 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
...
+
+		for scanner.Scan() {
+			if line >= start && line <= end {
+				text = append(text, scanner.Bytes()...)
+				text = append(text, "\n"...)
+			}
+			line++
+		}
+		if err := scanner.Err(); err != nil {
+			return nil, errors.Errorf("scanner couldn't read: %v", err)
```
bwplotka, line 162:
```suggestion
			return nil, errors.Wrap(err, "scanner read")
```

pkg/mdformatter/mdgen/mdgen_test.go
```diff
@@ -8,12 +8,40 @@ import (
...
+	}
+	
+	func CompareBytes(a []byte, aFn string, b []byte, bFn string) Diff {
+		return Compare(yoloString(a), aFn, yoloString(b), bFn)
+	}	
+	`
+)
+
 func TestFormat_FormatSingle_CodeBlockTransformer(t *testing.T) {
 	f := mdformatter.New(context.Background(), mdformatter.WithCodeBlockTransformer(NewCodeBlockTransformer()))
```
bwplotka, line 46: Why not putting this in our e2e framework with `input` `output` markdown files? (:

bwplotka (approved):
Nice, almost there 🤗

bwplotka (changes requested):
Just noticed 🙈 

To make sure we have one way of doing things... what's stops us without this PR to have mdox doing the same with:

````
```go mdox-gen-exec="sed -n '3,6p' main.go"
```
````

🤔 

While this PR is great, I fear it would be best to decouple this logic and allow better tools `sed` to be used 🙈 WDYT?

IF you agree, we could add test case with sed to ensure this works instead of full code.

saswatamcode:
Sure! Will close this and open a new PR with the sed test case!

PR #20, opened 2021-03-29
Add quotes to frontmatter

saswatamcode:
Currently, mdox removes all quotes from frontmatter values. But in case a value contains a semicolon, there might be yaml parsing issues in framworks like hugo. An example of this can be found [here](https://github.com/thanos-io/thanos/blob/main/docs/components/_index.md).
Before, this example would be outputted as,
```
title: Components:
```
This fix ensures that if a value has a semicolon, quotes would be preserved like so, 
```
title: "Components:"
```

### DIFF
--- a/pkg/mdformatter/mdformatter.go
+++ b/pkg/mdformatter/mdformatter.go
@@ -11,7 +11,7 @@ import (
 	"io/ioutil"
 	"os"
 	"sort"
-	"strings"
+	"strconv"
 
 	"github.com/Kunde21/markdownfmt/v2/markdown"
 	"github.com/bwplotka/mdox/pkg/gitdiff"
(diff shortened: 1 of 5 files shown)


pkg/mdformatter/mdformatter.go
```diff
@@ -117,10 +118,18 @@ func (FormatFrontMatter) TransformFrontMatter(_ SourceContext, frontMatter map[s
 			// Loop through all nested keys.
 			_, _ = fmt.Fprintf(b, "\n%v:", k)
 			for key, val := range frontMatterMap {
+				if v, ok := val.(string); ok && strings.Contains(v, ":") {
```
bwplotka, line 121: Suggestion: check if string and then always use `strconv.Quote` function? 🤗

bwplotka (commented):
Looks good! Can we have that tested too? 🤗 

Also, I would quote all strings then! (Consistency!)

bwplotka (approved):
Great! Thanks 💪🏽

PR #25, opened 2021-05-13
Add `sed` testcase

### LINKED ISSUE #18: Add `mdox-gen-file` to allow fetching from file

saswatamcode:
This PR adds in usage of `sed` with `mdox-gen-exec` as a testcase and mentions this in README as it's a common use case, as discussed in #19. 🙂
Resolves #18.

### DIFF
--- a/pkg/mdformatter/mdgen/testdata/mdgen_not_formatted.md
+++ b/pkg/mdformatter/mdgen/testdata/mdgen_not_formatted.md
@@ -38,3 +38,6 @@ alertmanagers:
 ```bash mdox-expect-exit-code=2 mdox-gen-exec="bash ./testdata/out3.sh"
 abc
 ```
+
+```bash mdox-gen-exec="sed -n '1,3p' ./testdata/out3.sh"
+```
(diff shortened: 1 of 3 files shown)


pkg/mdformatter/mdgen/testdata/mdgen_not_formatted.md
```diff
@@ -37,4 +37,7 @@ alertmanagers:
 
 ```bash mdox-expect-exit-code=2 mdox-gen-exec="bash ./testdata/out3.sh"
 abc
-```
\ No newline at end of file
+```
+
+```bash mdox-gen-exec="sed -n '1,3p' ./testdata/out3.sh"
```
bwplotka, line 42: Since we actually need only exec, wonder, do you think just `mdox=` or `mdox-exec=` would do the work? No need for `gen` then? (:

bwplotka (approved):
It works! 🙈 LGTM, just yolo idea.

PR #26, opened 2021-05-13
Fix same link in different file error

saswatamcode:
Currently, if there are same links but in different files (come under same file glob passed to mdox) like [here](https://github.com/thanos-io/thanos/blame/main/docs/contributing/mentorship.md#L26) and [here](https://github.com/thanos-io/thanos/blame/main/docs/contributing/community.md#L36),  mdox throws the error below,
```
remote link https://slack.cncf.io/: URL already visited
```
Even though the link is valid.
This PR fixes this and adds a test case for the same.

### DIFF
--- a/pkg/mdformatter/linktransformer/link.go
+++ b/pkg/mdformatter/linktransformer/link.go
@@ -150,15 +150,20 @@ func NewValidator(logger log.Logger, except *regexp.Regexp, anchorDir string) (m
 	}); err != nil {
 		return nil, err
 	}
+	v.c.OnRequest(func(request *colly.Request) {
+		v.rMu.Lock()
+		defer v.rMu.Unlock()
+		request.Ctx.Put("originalURL", request.URL.String())
+	})
 	v.c.OnScraped(func(response *colly.Response) {
 		v.rMu.Lock()
 		defer v.rMu.Unlock()
-		v.remoteLinks[response.Request.URL.String()] = nil
+		v.remoteLinks[response.Ctx.Get("originalURL")] = nil
 	})
 	v.c.OnError(func(response *colly.Response, err error) {
 		v.rMu.Lock()
 		defer v.rMu.Unlock()
-		v.remoteLinks[response.Request.URL.String()] = errors.Wrapf(err, "%q not accessible; status code %v", response.Request.URL.String(), response.StatusCode)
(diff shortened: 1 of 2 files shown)


saswatamcode:
Linter failure seems to be due to goreportcard.com being down.

bwplotka:
You should be able to retry @saswatamcode since you are maintainer now!

bwplotka:
Hm, do you know why this error was created in the first place? (:

saswatamcode:
I think this error was created to improve performance and avoid duplicate requests in colly. But here it seems to cause this issue...is there a better way to handle this other than using `AllowURLRevisit`? (:

Also, goreportcard seems to still be down so check would fail if I rerun...

pkg/mdformatter/linktransformer/link.go
```diff
@@ -150,15 +150,20 @@ func NewValidator(logger log.Logger, except *regexp.Regexp, anchorDir string) (m
 	}); err != nil {
 		return nil, err
 	}
+	v.c.OnRequest(func(request *colly.Request) {
+		v.rMu.Lock()
+		defer v.rMu.Unlock()
+		request.Ctx.Put("originalURL", request.URL.String())
```
bwplotka, line 156: Nice!

Suggestion: Handcrafted strings like this can be prone to errors. Use constant to reuse such key. Also it does not need to be string to be key. It's recommended to use empty struct (less mem allocated) e.g

`var originalURLKey = struct{}{} `

bwplotka (approved):
Great, one suggestion, otherwise LGTM (:

pkg/mdformatter/linktransformer/link.go
saswatamcode: Sure!

saswatamcode: The `Get` and `Put` methods seem to only take in strings as highlighted [here](https://github.com/gocolly/colly/blob/master/context.go#L23). Should `originalURLKey` be passed in using `fmt.Sprintf` here? But that maybe wouldn’t help with memory as that would be the same as using a string…

bwplotka: Ah right, this is something library specific not Go context.. ok, let's do constant at least

bwplotka (approved):

PR #29, opened 2021-05-16
Added transform command.

bwplotka:


bwplotka:
PTAL @saswatamcode  (:

README.md
```diff
@@ -4,16 +4,23 @@
...
 ## Goals
 
 Allow projects to have self-updating up-to-date documentation available in both markdown (e.g readable from GitHub) and static HTML. Hosted in the same repository as code and integrated with Pull Requests CI, hosting CD and code generation.
 
 ## Features
 
-* Enhanced amd consistent formatting for markdown files, focused on `.md` code readability.
-* Auto generation of code block content based on `mdoc-exec` directives (see [#code-generation](#code-generation)).
+* Enhanced and consistent formatting for markdown files in [GFM](https://github.github.com/gfm/) format, focused on readability.
+* Auto generation of code block content based on `mdoc-exec` directives (see [#code-generation](#code-generation)). Useful for:
```
saswatamcode, line 16:
```suggestion
* Auto generation of code block content based on `mdox-exec` directives (see [#code-generation](#code-generation)). Useful for:
```

main.go
```diff
@@ -202,15 +203,11 @@ func validateAnchorDir(anchorDir string, files []string) (_ string, err error) {
...
-	cmd := app.Command("web", "Tools for generating static HTML website based on https://gohugo.io/ on every PR with preview")
-	genCmd := cmd.Command("gen", "Generate versioned docs")
-
-	_ = genCmd.Arg("files", "Markdown file(s) to process.").Required().ExistingFiles()
-
-	// TODO(bwplotka): Generate versioned docs used for Hugo. Open question: Hugo specific? How to adjust links? Should fmt and this be
-	// the same?
+func registerTransform(_ context.Context, app *extkingpin.App) {
+	cmd := app.Command("transform", "Transform markdown files in various ways. For example pre process markdown files to allow it for use for popular static HTML websites based on markdown source code and front matter options.")
+	cfg := cmd.Flag("config", "Path to the YAML file with spec defined in github.com/bwplotka/mdox/pkg/transform.Config").
```
saswatamcode, line 208: Maybe a section can be added to README about `transform` command, with sample config yaml, and that can be linked instead of struct? 🙂

pkg/transform/transform.go
```diff
@@ -0,0 +1,377 @@
...
+			if err := os.RemoveAll(d); err != nil {
+				return err
+			}
+		}
+		if err := os.MkdirAll(d, os.ModePerm); err != nil {
+			return err
+		}
+
+		if c.GitIgnored {
+			if err = ioutil.WriteFile(filepath.Join(d, ".gitignore"), []byte("*.*"), os.ModePerm); err != nil {
```
saswatamcode, line 52: I think gitignore can just be `*` instead of `*.*`. This would work better if there is non-md file without extension, for example `Procfile`. 🙂

pkg/transform/transform.go
```diff
@@ -0,0 +1,377 @@
...
+	newAbsRelPathByOldAbsRelPath map[string]string
+}
+
+func (r *relLinkTransformer) TransformDestination(ctx mdformatter.SourceContext, destination []byte) ([]byte, error) {
+	split := strings.Split(string(destination), "#")
+	dest := split[0]
+	if strings.Contains(dest, "://") || filepath.IsAbs(dest) || strings.HasPrefix(string(destination), "#") {
+		return destination, nil
+	}
+	fmt.Println(ctx.Filepath)
```
saswatamcode, line 192: Linter fails due to this.

pkg/transform/transform.go
```diff
@@ -0,0 +1,377 @@
...
+// Super hacky for now.
+func popFirstHeader(path string) (_ string, rest []byte, err error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return "", nil, err
+	}
+	defer errcapture.Close(&err, file, "close file")
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
```
saswatamcode, line 362: Can we check `scanner.Err()` after this loop?

saswatamcode (approved):
This looks really amazing✨! Just some suggestions.

main.go
bwplotka: Well, we could use mdox-exec with our tool we plan to build and do this at the point? =D Great idea!

pkg/transform/transform.go
bwplotka: 👍🏽

pkg/transform/transform.go
bwplotka: Very good point, I missed this.

Where did you learn so many stuff about golang already??? ;p

main.go
saswatamcode: Sure!

PR #32, opened 2021-05-19
Added support for modifying links in image URLs. Fixed linking when adding extra files from outside.

bwplotka:


bwplotka:
Validated here: https://github.com/rhobs/handbook/pull/8

pkg/mdformatter/mdformatter_test.go
```diff
@@ -52,6 +53,7 @@ func TestCheck_NoTransformers(t *testing.T) {
 
 	exp, err := ioutil.ReadFile("testdata/not_formatted.md.diff")
 	testutil.Ok(t, err)
+	fmt.Println(diff.String())
```
saswatamcode, line 56: Can this be removed? Diff gets printed here irrespective of test passing/failing.  🙂

saswatamcode (approved):
This looks awesome! ✨ Just one tiny suggestion.

pkg/mdformatter/mdformatter_test.go
bwplotka: ah forgot to change

Issue #30, opened 2021-05-18
Endless redirection in few links


saswatamcode:
Some links redirect endlessly but aren't necessarily broken. For example "https://www.envoyproxy.io/", which generates the following error in mdox,
```
"https://www.envoyproxy.io/" not accessible; status code 301: Moved Permanently
```

curl output for the same,
```
curl -i https://www.envoyproxy.io/
HTTP/2 301 
cache-control: public, max-age=0, must-revalidate
content-length: 42
content-security-policy: frame-ancestors 'self';
content-type: text/plain
date: Tue, 18 May 2021 04:32:10 GMT
x-nf-request-id: fd4710cb-689c-4c75-b8f6-808272b5732b
location: https://www.envoyproxy.io/
server: Netlify
age: 25713

Redirecting to https://www.envoyproxy.io/
```
This link is perfectly viewable in a browser. Some way of handling links like this would be great!

PR #33, opened 2021-05-20
Add `links.validate.config` flag

saswatamcode:
This PR adds in the `--links.validate.config` flag which allows users to pass in path to YAML config with composable regex matchers. 

A sample yaml config would be,
```
validators:
  - regex: 'bwplotka\/mdox'
    token: '$(GITHUB_TOKEN)'
    type: "github"
  - regex: '(^http[s]?:\/\/)(www\.)?(fakelink[0-9]\.com\/)'
    type: "ignore"
  - regex: 'mycustomdomain'
    type: "roundtrip"
```

Here, all links are checked with default type `ignore`, i.e, links are not visited. If type is set to `roundtrip`, then links matching provided regex are visited by colly. For type `github` only repo name regex is taken, which then directs mdox to not visit any GitHub issue/PR link of that repo. User can also provide a token(env var substitution) to tackle GitHub API rate limiting.

Also, we don't check if a particular number is an issue or a PR since GitHub can automatically redirect such links to the correct page, for example the link`https://github.com/bwplotka/mdox/pull/30` is actually an issue and will be redirected. So such links can be safely skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ Flags:
       --links.validate.without-address-regex=^$  
                                  If specified, all links will be validated,
                                  except those matching the given target address.
+      --links.validate.without-github-links=""  
+                                 If specified, all links will be validated,
+                                 except the GitHub links for PRs and issues of
+                                 the given repo.
 
 Args:
   <files>  Markdown file(s) to process.
--- a/README.md
+++ b/README.md
@@ -60,13 +60,15 @@ Flags:
                                  exists).Absolute path links will be converted
                                  to relative links to anchor dir as well.
   -l, --links.validate           If true, all links will be validated
-      --links.validate.without-address-regex=^$  
-                                 If specified, all links will be validated,
-                                 except those matching the given target address.
-      --links.validate.config=""  
+      --links.validate.config-file=<file-path>  
                                  Path to YAML file for skipping link check, with
                                  spec defined in
-                                 github.com/bwplotka/mdox/pkg/linktranformer.Config
+                                 github.com/bwplotka/mdox/pkg/linktransformer.Config
+      --links.validate.config=<content>  
+                                 Alternative to 'links.validate.config-file'
+                                 flag (mutually exclusive). Content of YAML file
+                                 for skipping link check, with spec defined in
+                                 github.com/bwplotka/mdox/pkg/linktransformer.Config
 
 Args:
   <files>  Markdown file(s) to process.
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ Flags:
       --links.validate.without-address-regex=^$  
                                  If specified, all links will be validated,
                                  except those matching the given target address.
-      --links.validate.without-github-links=""  
-                                 If specified, all links will be validated,
-                                 except the GitHub links for PRs and issues of
-                                 the given repo.
+      --links.validate.config=""  
+                                 Path to YAML file for skipping link check, with
+                                 spec defined in
+                                 github.com/bwplotka/mdox/pkg/linktranformer.Config
 
 Args:
   <files>  Markdown file(s) to process.
--- a/main.go
+++ b/main.go
@@ -127,7 +127,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 		"Absolute path links will be converted to relative links to anchor dir as well.").Regexp()
 	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
-	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config", false)
+	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config", false, true)
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var opts []mdformatter.Option
--- a/main.go
+++ b/main.go
@@ -127,6 +127,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateExceptDomains := cmd.Flag("links.validate.without-address-regex", "If specified, all links will be validated, except those matching the given target address.").Default(`^$`).Regexp()
+	linksSkipGitHub := cmd.Flag("links.validate.without-github-links", "If specified, all links will be validated, except the GitHub links for PRs and issues of the given repo.").Default("").String()
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var opts []mdformatter.Option
@@ -151,7 +152,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 
 		var linkTr []mdformatter.LinkTransformer
 		if *linksValidateEnabled {
-			v, err := linktransformer.NewValidator(logger, *linksValidateExceptDomains, anchorDir)
+			v, err := linktransformer.NewValidator(logger, *linksValidateExceptDomains, *linksSkipGitHub, anchorDir)
 			if err != nil {
 				return err
 			}
--- a/main.go
+++ b/main.go
@@ -13,6 +13,7 @@ import (
 	"syscall"
 
 	"github.com/bwplotka/mdox/pkg/clilog"
+	"github.com/bwplotka/mdox/pkg/extflag"
 	"github.com/bwplotka/mdox/pkg/extkingpin"
 	"github.com/bwplotka/mdox/pkg/mdformatter"
 	"github.com/bwplotka/mdox/pkg/mdformatter/linktransformer"
@@ -126,8 +127,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 		"Absolute path links will be converted to relative links to anchor dir as well.").Regexp()
 	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
-	linksValidateExceptDomains := cmd.Flag("links.validate.without-address-regex", "If specified, all links will be validated, except those matching the given target address.").Default(`^$`).Regexp()
-	linksValidateConfig := cmd.Flag("links.validate.config", "Path to YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktranformer.Config").Default("").String()
+	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config", false)
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var opts []mdformatter.Option
@@ -152,7 +152,11 @@ This directive runs executable with arguments and put its stderr and stdout outp
 
 		var linkTr []mdformatter.LinkTransformer
 		if *linksValidateEnabled {
-			v, err := linktransformer.NewValidator(logger, *linksValidateExceptDomains, *linksValidateConfig, anchorDir)
+			validateConfigContent, err := linksValidateConfig.Content()
+			if err != nil {
+				return err
+			}
+			v, err := linktransformer.NewValidator(logger, validateConfigContent, anchorDir)
 			if err != nil {
 				return err
 			}
--- a/main.go
+++ b/main.go
@@ -127,7 +127,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateExceptDomains := cmd.Flag("links.validate.without-address-regex", "If specified, all links will be validated, except those matching the given target address.").Default(`^$`).Regexp()
-	linksSkipGitHub := cmd.Flag("links.validate.without-github-links", "If specified, all links will be validated, except the GitHub links for PRs and issues of the given repo.").Default("").String()
+	linksValidateConfig := cmd.Flag("links.validate.config", "Path to YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktranformer.Config").Default("").String()
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var opts []mdformatter.Option
@@ -152,7 +152,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 
 		var linkTr []mdformatter.LinkTransformer
 		if *linksValidateEnabled {
-			v, err := linktransformer.NewValidator(logger, *linksValidateExceptDomains, *linksSkipGitHub, anchorDir)
+			v, err := linktransformer.NewValidator(logger, *linksValidateExceptDomains, *linksValidateConfig, anchorDir)
 			if err != nil {
 				return err
 			}
--- a/main.go
+++ b/main.go
@@ -127,7 +127,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 		"Absolute path links will be converted to relative links to anchor dir as well.").Regexp()
 	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
-	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config", false, true)
+	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config", extflag.WithEnvSubstitution(true))
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var opts []mdformatter.Option
--- a/main.go
+++ b/main.go
@@ -127,7 +127,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 		"Absolute path links will be converted to relative links to anchor dir as well.").Regexp()
 	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
-	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config", extflag.WithEnvSubstitution(true))
+	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config", extflag.WithEnvSubstitution())
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var opts []mdformatter.Option
--- a/pkg/extflag/pathorcontent.go
+++ b/pkg/extflag/pathorcontent.go
@@ -10,6 +10,9 @@ package extflag
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
+	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 	"gopkg.in/alecthomas/kingpin.v2"
@@ -19,7 +22,8 @@ import (
 type PathOrContent struct {
 	flagName string
 
-	required bool
+	envSubstitution bool
+	required        bool
 
 	path    *string
 	content *string
@@ -30,7 +34,7 @@ type FlagClause interface {
 }
 
 // RegisterPathOrContent registers PathOrContent flag in kingpinCmdClause.
-func RegisterPathOrContent(cmd FlagClause, flagName string, help string, required bool) *PathOrContent {
+func RegisterPathOrContent(cmd FlagClause, flagName string, help string, required bool, envSubstitution bool) *PathOrContent {
 	fileFlagName := fmt.Sprintf("%s-file", flagName)
 	contentFlagName := flagName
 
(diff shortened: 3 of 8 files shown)


README.md
```diff
@@ -63,6 +63,10 @@ Flags:
       --links.validate.without-address-regex=^$  
                                  If specified, all links will be validated,
                                  except those matching the given target address.
+      --links.validate.without-github-links=""
```
bwplotka, line 66: Hm, this is great, but I wonder if we can introduce better composability here. 

E.g what if we find good mechanism for understanding GH wiki links, notifications or releases?

What if someone has website X and knows that links like a/b/c will work there as long as host is ok? 

I think we should allow some flexibility for users to add those, WDYT?

I would suggest something like:

```
--links.validate.config="<yaml content of filepath>"
```
(`pathorcontent` flag) with yaml like:

```
validate:
  validators:
     - regex: "(^http[s]?:\/\/)(www\.)github\.com"
       type: "github"
     # by default things are checked with `type: roundtrip`

```

.. so similar flow as with `globs` and transform yaml WDYT?

pkg/mdformatter/linktransformer/link.go
```diff
@@ -172,9 +186,59 @@ func NewValidator(logger log.Logger, except *regexp.Regexp, anchorDir string) (m
 	return v, nil
 }
 
+type GitHubResponse struct {
+	Number int `json:"number"`
+}
+
+func getGitHubRegex(reponame string) (*regexp.Regexp, int, error) {
```
bwplotka, line 193: Let's make it a separate validator, totally separate file etc as this is complex. For readability it's best to isolate complexities from each other and abstract them away (:

bwplotka (changes requested):
Amazing work. I would aim however for something more flexible, so we can add other smart "validators"  WDYT? (:

README.md
saswatamcode: I think that would be great! It would be better to allow users to pass in yaml config and maybe have multiple validators. Will try and implement this!

pkg/mdformatter/linktransformer/link.go
saswatamcode: Sure!

README.md
```diff
@@ -63,6 +63,10 @@ Flags:
       --links.validate.without-address-regex=^$  
                                  If specified, all links will be validated,
                                  except those matching the given target address.
+      --links.validate.config=""  
+                                 Path to YAML file for skipping link check, with
+                                 spec defined in
+                                 github.com/bwplotka/mdox/pkg/linktranformer.Config
```
bwplotka, line 69:
```suggestion
                                 github.com/bwplotka/mdox/pkg/linktransformer.Config
```

README.md
```diff
@@ -63,6 +63,10 @@ Flags:
       --links.validate.without-address-regex=^$
```
bwplotka, line 63: Let's kill this flag and use below one maybe?

(We can adjust flag names before first release - later it will be pain for users so let's make sure we got it right now)

main.go
```diff
@@ -127,6 +127,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateExceptDomains := cmd.Flag("links.validate.without-address-regex", "If specified, all links will be validated, except those matching the given target address.").Default(`^$`).Regexp()
+	linksValidateConfig := cmd.Flag("links.validate.config", "Path to YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktranformer.Config").Default("").String()
```
bwplotka, line 130:
```suggestion
	linksValidateConfig := cmd.Flag("links.validate.config", "Path to YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.Config").Default("").String()
```

bwplotka, line 130: Can we use `pathorcontent` flag?

pkg/mdformatter/linktransformer/config.go
```diff
@@ -0,0 +1,64 @@
...
+	"regexp"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+type Config struct {
+	Version int
+
+	Validate struct {
```
bwplotka, line 19: Why not stright `Validators` field here?

pkg/mdformatter/linktransformer/config.go
```diff
@@ -0,0 +1,64 @@
...
+		Validators []Validator `yaml:"validators"`
+	}
+}
+
+type Validator struct {
+	_regex  *regexp.Regexp
+	_maxnum int
```
bwplotka, line 26:
```suggestion
	_maxNum int
```

pkg/mdformatter/linktransformer/config.go
```diff
@@ -0,0 +1,64 @@
...
+	// Regex for type github is reponame matcher, like `bwplotka\/mdox`.
```
bwplotka, line 27: But what regex is for? Can we encode it in comment and field name?

bwplotka: What about making a similar format as mdox web transformer config? So something like 

```
validators:
- glob: ..
  type: "github"    
```

bwplotka: (first match wins)

pkg/mdformatter/linktransformer/link.go
```diff
@@ -169,12 +175,16 @@ func NewValidator(logger log.Logger, except *regexp.Regexp, anchorDir string) (m
...
+	if err != nil {
+		return nil, err
+	}
 	return v, nil
 }
 
 // MustNewValidator returns mdformatter.LinkTransformer that crawls all links.
-func MustNewValidator(logger log.Logger, except *regexp.Regexp, anchorDir string) mdformatter.LinkTransformer {
-	v, err := NewValidator(logger, except, anchorDir)
+func MustNewValidator(logger log.Logger, except *regexp.Regexp, linksValidateConfig string, anchorDir string) mdformatter.LinkTransformer {
```
bwplotka, line 186: I feel it would be best to get parsed Config here WDYT?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,112 @@
...
+const (
+	gitHubAPIURL = "https://api.github.com/repos/%v/%v?sort=created&direction=desc&per_page=1"
+)
+
+type GitHubResponse struct {
+	Number int `json:"number"`
+}
+
+// Match link with any one of provided validators.
+func CheckValidators(dest string, v Config) bool {
```
bwplotka, line 27:
```suggestion
func (v Config).GetValidatorForURL(url string) SomeValidator interface {
```

Let's have some interface for each validator so it's cleaner, WDYT?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,112 @@
...
+				return false
+			}
+			return true
+		}
+	}
+	return false
+}
+
+// If type is "github", change regex and add maxnum.
+func CheckGitHub(v Config) error {
```
bwplotka, line 49: Does it need to be public? What check means? You mean `func (v Config) validateGH() error { `

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+	return nil
+}
+
+// Get GitHub pulls/issues regex from repo name.
+func getGitHubRegex(repoRe string) (*regexp.Regexp, int, error) {
+	if repoRe != "" {
+		// Get reponame from regex.
+		idx := strings.Index(repoRe, `\`)
```
bwplotka, line 67: Hm, it is bit confusing. What about either expecting named regex group? `(?P<repo>.*)

bwplotka (commented):
Nice, Looks quite good, some nits.

README.md
saswatamcode: Sure! Would be easier!

main.go
saswatamcode: Sure!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: Sure!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: I'll try to implement this!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: I'll try to implement this!

pkg/mdformatter/linktransformer/config.go
```diff
@@ -0,0 +1,48 @@
...
+	Validators []Validator `yaml:"validators"`
+}
+
+type Validator struct {
+	_regex  *regexp.Regexp
+	_maxNum int
+	// Regex for type github is reponame matcher, like `bwplotka\/mdox`.
+	Regex string `yaml:"regex"`
+	// By default type is `roundtrip`. Could be `github`.
+	Type string `yaml:"type"`
```
bwplotka, line 26: let's make it a `type ValidatorType string` and then provide 2 consts with those expected values (: It will be easier to use, at least in code level

pkg/mdformatter/linktransformer/config.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func ParseConfig(c []byte) (Config, error) {
+	if string(c) == "" {
```
bwplotka, line 30: This can surprise. Imagine someone will pass empty bytes by mistake and parse will just succeed. Can we do this on caller side?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,120 @@
...
+
+type URLValidator struct {
+	Matched bool
+}
+
+// Match link with any one of provided validators.
+func (v Config) GetValidatorForURL(url string) URLValidator {
+	u := URLValidator{Matched: false}
+	for _, val := range v.Validators {
+		if val._regex.MatchString(url) {
```
bwplotka, line 34: Let's invert. Prefer keep the complex logic with as small indent as possible. We found invert here and continue if not matched avoiding one indent. Same below.

See: https://thanos.io/tip/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,120 @@
...
+			if val.Type == "github" {
```
bwplotka, line 35: Let's use proposed const

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,120 @@
...
+				// Find rightmost index of match i.e, where regex match ends.
+				// This will be where issue/PR number starts. Split incase of section link and convert to int.
+				idx := val._regex.FindStringIndex(url)
```
bwplotka, line 38: idx?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,120 @@
...
+				stringNum := strings.Split(url[idx[1]:], "#")
+				num, err := strconv.Atoi(stringNum[0])
+				// If number in link does not exceed then link is valid. Otherwise will be checked by v.c.Visit.
+				if val._maxNum >= num && err == nil {
+					u.Matched = true
+					return u
```
bwplotka, line 44: Why return u if when this block ends also  return u?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,120 @@
...
+				return u
+			}
+			u.Matched = true
+			return u
+		}
+	}
+	return u
+}
+
+// If type is "github", change regex and add maxnum.
```
bwplotka, line 55: Comments should always start with the function name.

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,120 @@
...
+func (v Config) validateGH() error {
+	for i := range v.Validators {
+		if v.Validators[i].Type == "github" {
```
bwplotka, line 58: ditto - invert fo readability

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,120 @@
...
+			v.Validators[i]._regex = regex
+			v.Validators[i]._maxNum = maxNum
+		}
+	}
+	return nil
+}
+
+// Get GitHub pulls/issues regex from repo name.
+func getGitHubRegex(repoRe string) (*regexp.Regexp, int, error) {
+	if repoRe != "" {
```
bwplotka, line 72: ditto and please remember about this rule. If you invert this and improve readability a lot.

Also having one big IF at the beginning is a code smell. Why invoking this function in the first place if repoRe is empty? 🤔

bwplotka (commented):
Thanks a lot! Much better (: Some thoughts. Almost perfect!

pkg/mdformatter/linktransformer/config.go
saswatamcode: Sure!

pkg/mdformatter/linktransformer/config.go
saswatamcode: Sure! Currently, after inverting fixes, only one const is used(`roundtrip`) so, commenting the second one until it is used in future.

pkg/mdformatter/linktransformer/validator.go
saswatamcode: Alright! Have made the same mistake in a number of places. Will rectify!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: This stores the rightmost index of the match, i.e gives the index of an issue/PR number. Will change the name!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: I seem to have missed this. Will remove!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: Sure! Will keep this in mind in future PRs.

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,145 @@
...
+type GitHubResponse struct {
+	Number int `json:"number"`
+}
+
+type URLValidator struct {
+	Matched bool
+}
+
+// GetValidatorForURL matches link with any one of provided validators.
+func (v Config) GetValidatorForURL(url string) URLValidator {
```
bwplotka, line 32: We need to work on different abstrations here...

What about something else:

```
type Validator interface {
IsValid(URL string) (bool, error)
}
```

* Type == github, where we can have githubValidator struct, which does gh check
* Type  == rt rtValidator, which does normal look up 
image
* Type == static/ignore which is always true?

bwplotka (commented):
Some comments during our 1:1, great job so far!

pkg/extflag/pathorcontent.go
```diff
@@ -0,0 +1,103 @@
...
+	path    *string
+	content *string
+}
+
+type FlagClause interface {
+	Flag(name, help string) *kingpin.FlagClause
+}
+
+// RegisterPathOrContent registers PathOrContent flag in kingpinCmdClause.
+func RegisterPathOrContent(cmd FlagClause, flagName string, help string, required bool, envSubstitution bool) *PathOrContent {
```
bwplotka, line 37: Time for variadic options here (: 

```suggestion
func RegisterPathOrContent(cmd FlagClause, flagName string, help string, opts ...Option) *PathOrContent {
```

pkg/extflag/pathorcontent.go
```diff
@@ -0,0 +1,103 @@
...
+	if p.envSubstitution {
+		content = substituteEnvVars(string(content))
+	}
+	return content, nil
+}
+
+// substituteEnvVars returns content of YAML file with substituted environment variables.
+// Will be substituted with empty string if env var isn't set.
+// Follows K8s convention, i.e $(...), as mentioned here https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/.
+func substituteEnvVars(content string) []byte {
```
bwplotka, line 91: Let's take in bytes if we want to split out bytes (consistency)

bwplotka: Can we reuse exsiting code? https://github.com/thanos-io/thanos/blob/7d118f416be98cf502b9b5e0df4e98b37478f3f3/pkg/reloader/reloader.go#L419

bwplotka (commented):
That looks solid to me (:

pkg/extflag/pathorcontent.go
saswatamcode: Sure!

pkg/extflag/pathorcontent.go
saswatamcode: Alright! Will try to reuse code from here!

saswatamcode: Have taken existing code but used `os.Getenv` instead of `os.LookupEnv` since `Token` here is optional and don't want to error out if variable is unset. 🙂

bwplotka: Why we don't want to error out? I think that is a correct behaviour

saswatamcode: The `GITHUB_TOKEN` variable is only expected to be set in CI runners so that authenticated API calls are made and it's not rate limited(Same IP across macos runners problem). But if user decides not to set the token i.e, isn't getting rate limited and doesn't need to make auth'd requests then there shouldn't be any error I think. So should I put the error back in?

bwplotka: >  But if user decides not to set the token i.e, isn't getting rate limited and doesn't need to make auth'd requests then there shouldn't be any error I think. 

If user decided not to set this token, why adding this to yaml as `$(....)` construct?

saswatamcode: If they wanted to run CI job locally but then they can have a make-docs-local command as well. I see what you mean, I'll add error back in! 🙂

bwplotka: explicit changes are better than surprising ones

pkg/extflag/pathorcontent.go
```diff
@@ -0,0 +1,130 @@
...
+
+// WithRequired allows you to override default required option.
+func WithRequired(r bool) Option {
+	return func(p *PathOrContent) {
+		p.required = r
+	}
+}
+
+// WithRequired allows you to override default envSubstitution option.
+func WithEnvSubstitution(e bool) Option {
```
bwplotka, line 106:
```suggestion
func WithEnvSubstitution() Option {
```

pkg/extflag/pathorcontent.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			return nil, err
+		}
+		content = replace
+	}
+	return content, nil
+}
+
+// WithRequired allows you to override default required option.
+func WithRequired(r bool) Option {
```
bwplotka, line 99:
```suggestion
func WithRequired() Option {
```

bwplotka: we don't need to expose both bools, as we have by default false, so we need to have just method for true, right?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	// If number in link does not exceed then link is valid.
+	if v._maxNum >= num {
+		return true, nil
+	}
+	return false, nil
+}
+
+// RoundTripValidator.IsValid returns false if url matches, to ensure it is visited by colly.
+func (v RoundTripValidator) IsValid(URL string) (bool, error) {
```
bwplotka, line 33: This still needs some love 🙈 

"RountTripValidator says URL is invalid" - that does not sounds like this would mean it will be check by colly. I would rather put colly logic directly here into this method, no?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,76 @@
...
+// GetValidatorForURL returns correct Validator by matching URL.
+func (v Config) GetValidatorForURL(URL string) Validator {
+	var u Validator
+	for _, val := range v.Validators {
+		switch val.Type {
+		case roundtripValidator:
+			if !val.rtValidator._regex.MatchString(URL) {
+				continue
+			}
+			u = val.rtValidator
```
bwplotka, line 51: We need `u` variable, we can just return, no?

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+			u = val.ghValidator
+			return u
+		case ignoreValidator:
+			if !val.igValidator._regex.MatchString(URL) {
+				continue
+			}
+			u = val.igValidator
+			return u
+		default:
```
bwplotka, line 65: We don't expect anything, so we can just panic for safety. It's ok to panic for totally impossible things, that might surprise us some day (as being suddenly possible)

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,76 @@
...
+			continue
+		}
+	}
+	// By default all links are ignored.
+	u = IgnoreValidator{}
```
bwplotka, line 70: Hm... I don't think, I think we want to round trip rest?

bwplotka (commented):
Thanks a lot! Super close to merge

pkg/extflag/pathorcontent.go
saswatamcode: Yes. Will change this!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: Alright, will put this logic directly! 🙂

pkg/mdformatter/linktransformer/validator.go
saswatamcode: Sure! I missed this, will rectify!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: Okay, I might've misunderstood earlier. Will change!

pkg/mdformatter/linktransformer/validator.go
saswatamcode: Oh okay, then will replace with panic.

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,89 @@
...
+	if _, ok := r.remoteLinks[k.dest]; ok {
+		r.rMu.RUnlock()
+		return true, nil
+	}
+	r.rMu.RUnlock()
+
+	r.rMu.Lock()
+	defer r.rMu.Unlock()
+	// We need to check again here to avoid race.
+	if _, ok := r.remoteLinks[k.dest]; ok {
```
bwplotka, line 48: This is some sort of "caching" that we can use on TOP of validators, so other validators can leverage it no? It's ok to think about it later PRs (:

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -0,0 +1,89 @@
...
+				continue
+			}
+			return val.ghValidator
+		case ignoreValidator:
+			if !val.igValidator._regex.MatchString(URL) {
+				continue
+			}
+			return val.igValidator
+		default:
+			panic("unexpected validator type")
```
bwplotka, line 84:
```suggestion
			panic(fmt.Sprintf("unexpected validator type %v", val.Type))
```

bwplotka (approved):
Thanks! LGTM (: Two issues, but it's ok to fix in later PRs 💪🏽

PR #37, opened 2021-06-07
Unify config parsers

saswatamcode:
This PR unifies the config parser of `transform --config` and `fmt --links.validate.config` flag into one package and changes `transform --config` flag to `pathorcontent`. 

Currently, two separate files need to be provided each for transform and link validation (even if two YAML docs are specified in one file, it will fail based on the order of docs). With this change, it will allow user to specify both transform and link validation config in one YAML file.


pkg/config/config.go
```diff
@@ -48,10 +57,13 @@ type Config struct {
...
 	// LocalLinksStyle sets linking style to be applied. If empty, we assume default style.
 	LocalLinksStyle LocalLinksStyle `yaml:"localLinksStyle"`
+
+	// Validators set multiple composable validators for link checking.
+	Validators []ValidatorConfig `yaml:"validators"`
 }
 
 type TransformationConfig struct {
-	_glob glob.Glob
+	ParsedGlob glob.Glob
```
bwplotka, line 66:
```suggestion
	parsedGlob glob.Glob
```
If you make it public it will be parsable by yaml decoders (and other decoders )

bwplotka (changes requested):
Hmmmmmmm I am not sure of this idea.

I would keep transform and validators in separate package configs really. Those are totally orthogonal things to configure, no?

pkg/config/config.go
saswatamcode: Sure! Also, these are both totally orthogonal things but as it is currently, a user would have to specify two different YAML files(if they choose not to go with just passing content) to make both transform and format work. Would that be okay? Then I can close this PR. 🙂

bwplotka: Yes, why not? Those are two separate logics (:

bwplotka: Two different commands even.

saswatamcode: Okay then! 🙂

PR #40, opened 2021-06-11
Add retries to link checking

### LINKED ISSUE #11: Add retries to link checking.
error: 		"http://semver.org/spec/v2.0.0.html" not accessible; status code 503: Service Unavailable

saswatamcode:
This PR adds retries to link checking based on status code. For status code 429, it uses the `Retry-After` header to wait before retrying and retries normally for status code 301, 307 and 503. Can try a maximum of three times before returning error. This can increase the time of processing docs if rate limit is encountered but not too excessively.
Fixes #11.

Some questions:
- Should this logic be put behind a `--retry` flag so that user can en/disable and set number of retries?
- The `Retry-After` header works well for sites like GitHub which returns 429 with `Retry-After` set to 60s. But other sites might make value excessive, so should a limit be placed?


pkg/mdformatter/linktransformer/link.go
```diff
@@ -175,7 +178,30 @@ func NewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir strin
...
-		v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(err, "%q not accessible; status code %v", response.Request.URL.String(), response.StatusCode)
+		retriesStr := response.Ctx.Get(retryKey)
+		retries, _ := strconv.Atoi(retriesStr)
+		switch response.StatusCode {
+		case 429:
+			if retries <= 2 {
+				response.Ctx.Put(retryKey, strconv.Itoa(retries+1))
+				retryAfter, convErr := strconv.Atoi(response.Headers.Get("Retry-After"))
+				if convErr == nil {
+					time.Sleep(time.Duration(retryAfter) * time.Second)
```
bwplotka, line 189: hmmmmm I think we need to have global retry. You see that will not work if others for same host will try in the same time 🤔

pkg/mdformatter/linktransformer/link.go
```diff
@@ -175,7 +178,30 @@ func NewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir strin
 	v.c.OnError(func(response *colly.Response, err error) {
 		v.rMu.Lock()
 		defer v.rMu.Unlock()
-		v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(err, "%q not accessible; status code %v", response.Request.URL.String(), response.StatusCode)
+		retriesStr := response.Ctx.Get(retryKey)
+		retries, _ := strconv.Atoi(retriesStr)
+		switch response.StatusCode {
+		case 429:
+			if retries <= 2 {
+				response.Ctx.Put(retryKey, strconv.Itoa(retries+1))
```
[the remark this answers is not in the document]
bwplotka: Is it because `.Retry()`  trigger this OnError too?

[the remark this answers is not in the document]
bwplotka: Can we add comment about it?

pkg/mdformatter/linktransformer/link.go
```diff
@@ -175,7 +178,30 @@ func NewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir strin
...
+				retryAfter, convErr := strconv.Atoi(response.Headers.Get("Retry-After"))
+				if convErr == nil {
+					time.Sleep(time.Duration(retryAfter) * time.Second)
+				}
+
+				retryErr := response.Request.Retry()
```
bwplotka, line 192: I might think we need default back of maybe?

pkg/mdformatter/linktransformer/link.go
saswatamcode: Oh! I didn't think of this before! Will try to implement global retry!

pkg/mdformatter/linktransformer/link.go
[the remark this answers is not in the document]
saswatamcode: Yes! It triggers same methods as it calls same function as `Visit()` (highlighted [here](https://github.com/gocolly/colly/blob/master/request.go#L151) and [here](https://github.com/gocolly/colly/blob/master/request.go#L115)). So we increment retryKey via this. Will add in a comment!

pkg/mdformatter/linktransformer/link.go
```diff
@@ -175,7 +178,30 @@ func NewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir strin
...
+				v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(retryErr, "%q rate limited even after retry; status code %v", response.Request.URL.String(), response.StatusCode)
+			}
+		case 301, 307, 503:
+			if retries <= 2 {
```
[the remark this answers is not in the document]
saswatamcode: I set it to this in kind of yolo manner since I wasn't sure what a good default value would be. I was thinking this could be replaced by an option/flag later on!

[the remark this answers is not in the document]
saswatamcode: I set it to this in kind of yolo manner since I wasn't sure what a good default value would be. I was thinking this could be replaced by an option/flag later on!

pkg/mdformatter/linktransformer/link.go
bwplotka: well default if no retry-after is set? maybe something like 1s

pkg/mdformatter/linktransformer/link.go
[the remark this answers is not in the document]
bwplotka: so what happens if we retry too many times? II can see we just return `v,nil` meaning all ok?

pkg/mdformatter/linktransformer/link.go
saswatamcode: Oh okay! Will try to implement this!

pkg/mdformatter/linktransformer/link.go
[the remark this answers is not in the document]
saswatamcode: A limit needs to be set so that it doesn't retry infinitely or too many times so I went with three but even if we did retry too many times there possibly wouldn't be any error just the statuscode would remain the same and take a long time. So I didn't return any error from there.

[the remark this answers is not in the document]
saswatamcode: Another maybe better approach could be retrying once and seeing if the result is the same as the previous statuscode, then error out. So we only retry once by default...

pkg/mdformatter/linktransformer/link.go
saswatamcode: I thought about this and tested it a bit. I think we never make requests to the same host again or at the same time due to our present caching layer and retries can only occur after request! So this should work. 🙂

bwplotka: I challenge this statement (:

bwplotka: I think it's fine for now, let's iterate (global retry will  be hard to implement), but I think it's needed, paths are cached not hosts

pkg/mdformatter/linktransformer/link.go
```diff
@@ -175,7 +179,40 @@ func NewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir strin
 	v.c.OnError(func(response *colly.Response, err error) {
 		v.rMu.Lock()
 		defer v.rMu.Unlock()
-		v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(err, "%q not accessible; status code %v", response.Request.URL.String(), response.StatusCode)
+		retriesStr := response.Ctx.Get(numberOfRetriesKey)
+		retries, _ := strconv.Atoi(retriesStr)
+		switch response.StatusCode {
+		case http.StatusTooManyRequests:
+			if retries <= 0 {
```
bwplotka, line 186: ditto (indent)

bwplotka (commented):
Some review during our 1:1

pkg/mdformatter/linktransformer/link.go
```diff
@@ -175,7 +179,40 @@ func NewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir strin
...
+				var retryAfter int
+				// Retry calls same methods as Visit and makes request with same options.
+				// So retryKey is incremented here if onError is called again after Retry. By default retries once.
+				response.Ctx.Put(numberOfRetriesKey, strconv.Itoa(retries+1))
+				retryAfter, convErr := strconv.Atoi(response.Headers.Get("Retry-After"))
+				if convErr != nil {
+					retryAfter = 1
+				}
+				time.Sleep(time.Duration(retryAfter) * time.Second)
```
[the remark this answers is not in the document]
saswatamcode: Colly allows us to pass in context and use that, but it isn't released yet. So pinning to needed commit for now! 🙂https://pkg.go.dev/github.com/gocolly/colly/v2@v2.1.1-0.20201013153555-8252c346cfb0#Collector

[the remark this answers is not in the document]
bwplotka: perfect

pkg/mdformatter/linktransformer/link.go
```diff
@@ -175,14 +180,53 @@ func NewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir strin
...
+			v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(err, "%q not accessible; status code %v", response.Request.URL.String(), response.StatusCode)
+		}
 	})
 	return v, nil
 }
 
 // MustNewValidator returns mdformatter.LinkTransformer that crawls all links.
 func MustNewValidator(logger log.Logger, linksValidateConfig []byte, anchorDir string) mdformatter.LinkTransformer {
-	v, err := NewValidator(logger, linksValidateConfig, anchorDir)
+	v, err := NewValidator(context.TODO(), logger, linksValidateConfig, anchorDir)
```
bwplotka, line 229: does not sound useful ;p

bwplotka (approved):
I think retrying will be annoying if we cannot interrupt it. Do you mind hooking the context properly. It will also show if we put it in right place... (:

pkg/mdformatter/linktransformer/link.go
saswatamcode: We can currently interrupt retry. For instance, testing a file with envoyproxy link and setting retryAfterSeconds to 10, gives the following results. First is when I let it run and retry, second is when I interrupt,

```
time mdox fmt -l test.md
error:          test.md: "https://envoyproxy.io" not accessible even after retry; status code 301: Moved Permanently
mdox fmt -l test.md  0.14s user 0.06s system 1% cpu 14.995 total

ime mdox fmt -l test.md
^Cinfo: caught signal. Exiting.: interrupt
mdox fmt -l test.md  0.12s user 0.05s system 2% cpu 6.762 total
```
So we can cancel retry without having to wait. But I'm not sure what would be the correct way to hook context?

saswatamcode: Yes, we pass it [here](https://github.com/bwplotka/mdox/blob/main/pkg/mdformatter/linktransformer/link.go#L158)! That's why we can cancel without waiting.

Issue #23, opened 2021-05-12
Add subcommand that will generate example YAML that will parsable from Go structs


bwplotka:
Similar to https://github.com/thanos-io/thanos/blob/main/scripts/cfggen/main.go

Acceptance Criteria:

* User can pass such tool in `mdox-exec` feature (separate subcommand?)
* Will generate example YAML (or whatever format is parsed to Go structs)
* Allows passing example defaults
* Preserve comments
* Optional: Tests if generated yaml can be actually parsable and returns err if not?

PR #41, opened 2021-06-11
Fixed mdox gen errors on command with `=` inside (common case).

bwplotka:

### DIFF
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -5,7 +5,6 @@ package mdgen
 
 import (
 	"bytes"
-	"context"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -17,8 +16,6 @@ import (
 )
 
 const (
-	infoStringKeyLang     = "mdox-gen-lang"
-	infoStringKeyType     = "mdox-gen-type"
 	infoStringKeyExec     = "mdox-exec"
 	infoStringKeyExitCode = "mdox-expect-exit-code"
 )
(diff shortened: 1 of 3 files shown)


pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -40,19 +37,22 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 	}
 	infoStringAttr := map[string]string{}
 	for i, field := range infoFiels {
-		val := strings.Split(field, "=")
+		val := []string{field}
+		if i := strings.Index(field, "="); i != -1 {
+			val = []string{field[:i], field[i+1:]}
```
saswatamcode, line 42: Awesome! Now it is only split at first "=". But maybe instead of using `Index` this can be done with something like,
```
val := strings.SplitN(field, "=", 2)
```
This also would only split at first "=" and return slice of length 2. WDYT?🙂

pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -82,30 +82,12 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
...
-		default:
-			return nil, errors.Errorf("expected language a first element of info string got %q; Got info string %q", lang, string(infoString))
-		}
-	}
 	panic("should never get here")
 }
 
 func (t *genCodeBlockTransformer) Close(ctx mdformatter.SourceContext) error { return nil }
-
-func genGo(ctx context.Context, moduleRoot string, typePath string) ([]byte, error) {
```
saswatamcode, line 108: I thought this was to be done as mentioned in #23. Do you have something else in mind? 🙂

pkg/mdformatter/mdgen/testdata/mdgen_formatted.md
```diff
@@ -40,3 +36,8 @@ test output3
 
 echo "test output3"
 ```
+
+```yaml mdox-exec="bash ./testdata/out2.sh --name=queryfrontend.InMemoryResponseCacheConfig"
```
saswatamcode, line 40: Nice testcase! 💪🏼

saswatamcode (approved):
This looks awesome! ✨ Just some suggestions!

pkg/mdformatter/mdgen/mdgen.go
bwplotka: Tried to use it, but found it confusing why 2 not 1.. (: will try thanks!

pkg/mdformatter/mdgen/mdgen.go
bwplotka: I think this part is not clearly understood yet. We don't know yet how it should look.

This snippet was mainly for the beginning to show you and Uche how we might want to do it.

Normally it's not the best to maintain code which is not used (YAGNI) so killed it (:

pkg/mdformatter/mdgen/testdata/mdgen_formatted.md
bwplotka: Yea. It was failing for us before

pkg/mdformatter/mdgen/mdgen.go
saswatamcode: Oh okay!

pkg/mdformatter/mdgen/mdgen.go
saswatamcode: I think 2 is the number of substrings returned...but yes it is not intuitive. This can be skipped then I guess.

Issue #42, opened 2021-06-11
Make sure README is up to date (check on CI), point to correct validator config.


bwplotka:
Correct struct `ValidatorConfig` not  what is pointed for `links.validate.config`

saswatamcode:
Closed via #44.

PR #44, opened 2021-06-13
Add CI for docs and fix dash header ID

### LINKED ISSUE #43: Markdown header IDs not parsed correctly
Markdown headers like `## Expose UI on a sub-path` get parsed as id `expose-ui-on-subpath` which isn't correct as dash gets removed. Correct id should be `expose-ui-on-a-sub-path` as demonstrated [here](https://github.com/thanos-io/thanos/blob/main/docs/components/query.md#expose-ui-on-a-sub-path).

saswatamcode:
This PR adds in CI for documentation check and make sure `links.validate.config` points to correct struct as mentioned in #42 . Also fixes markdown header ID issue and fixes #43.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,31 @@
+name: docs
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+  pull_request:
+
+jobs:
+  check:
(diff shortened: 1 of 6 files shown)


bwplotka (approved):
LGTM, but normally let's aim for decoupled PRs (:

PR #45, opened 2021-06-15
Change transform config flag to `pathorcontent`

saswatamcode:
This PR changes transform config flag to `pathorcontent`. A particular use case for this would be dynamic input and output dirs (for example when transforming docs for multiple releases of a project).


bwplotka (approved):
💪🏽 LGTM

PR #46, opened 2021-06-16
Added support for formatting in-md HTML and comments.

bwplotka:

### DIFF
--- a/pkg/mdformatter/transformer.go
+++ b/pkg/mdformatter/transformer.go
@@ -11,6 +11,7 @@ import (
 	"github.com/yuin/goldmark/ast"
 	"github.com/yuin/goldmark/renderer"
 	"github.com/yuin/goldmark/text"
+	"golang.org/x/net/html"
 )
 
 type nopOpsRenderer struct {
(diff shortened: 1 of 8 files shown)


pkg/mdformatter/transformer.go
```diff
@@ -40,7 +41,56 @@ func (t *transformer) Render(w io.Writer, source []byte, node ast.Node) error {
...
+		case *ast.HTMLBlock, *ast.RawHTML:
+			if !entering {
+				return ast.WalkSkipChildren, nil
+			}
+
+			// Parse HTML to get inline links on our own, goldmark does not do that.
+			b := bytes.Buffer{}
+			for i := 0; i < node.Lines().Len(); i++ {
+				segment := node.Lines().At(i)
+				_, _ = b.Write(segment.Value(source))
```
saswatamcode, line 53: Seems like node stays empty in this. So it is never tokenized and the first line of HTML block gets removed, which is also why `format not formatted` and `format formatted` tests both fail.

saswatamcode (approved):
Awesome work! 💫   Just some suggestions.

bwplotka:
Fixed all! (: @saswatamcode

PR #47, opened 2021-06-17
Add `backMatter` support

saswatamcode:
This PR adds support for "back matter" config in transform.

Sematics are similar to frontmatter but it is not parsed as YAML, only text.

For example,
```
    backMatter:
      template: |
        Found a typo, inconsistency or missing information in our docs?
        Help us to improve [Thanos](https://thanos.io) documentation by proposing a fix [on GitHub here](https://github.com/thanos-io/thanos/edit/main/docs/{{ .Origin.Filename }}) :heart:
```


pkg/transform/transform.go
```diff
@@ -323,6 +349,29 @@ func (f *frontMatterTransformer) TransformFrontMatter(ctx mdformatter.SourceCont
...
+	}); err != nil {
+		return nil, err
+	}
+
+	// Add demarkation to back matter.
+	demark := "\n\n---\n"
+	m := b.Bytes()
+	m = append([]byte(demark), m...)
+
+	// Back matter is not parsed as YAML since it is generally assumed to be a block of text.
```
bwplotka, line 369: This might be confusing that one is yaml, one not, maybe worth to document it better.

bwplotka (approved):
LGTM, just one nit, but let's fix later.

PR #52, opened 2021-06-24
Fix i18n section links

### LINKED ISSUE #50: Section links for md in non-English language break
Example:
https://github.com/kubernetes/website/blob/main/README-hi.md

saswatamcode:
This PR converts `toHeaderID` function to use a unicode regex to generate header ids for any given language instead of using ascii codes which restrict section links to just English. Test case added as well. Fixes #50.

### DIFF
--- a/pkg/mdformatter/linktransformer/link.go
+++ b/pkg/mdformatter/linktransformer/link.go
@@ -387,17 +387,24 @@ func (l localLinksCache) addRelLinks(localLink string) error {
 
 func toHeaderID(header []byte) string {
 	var id []byte
+	// Remove punctuation from header except '-' or '#'.
+	// '\p{L}\p{N}\p{M}' is the unicode equivalent of '\w', https://www.regular-expressions.info/unicode.html.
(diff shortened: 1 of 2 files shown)


pkg/mdformatter/linktransformer/link.go
```diff
@@ -387,17 +387,24 @@ func (l localLinksCache) addRelLinks(localLink string) error {
 
 func toHeaderID(header []byte) string {
 	var id []byte
+	// Remove punctuation from header except '-' or '#'.
+	// '\p{L}\p{N}\p{M}' is the unicode equivalent of '\w', https://www.regular-expressions.info/unicode.html.
+	punctuation := regexp.MustCompile(`[^\p{L}\p{N}\p{M}-# ]`)
+	header = punctuation.ReplaceAll(header, []byte(""))
+	headerText := bytes.TrimLeft(bytes.ToLower(header), "#")
+	// If header is just punctuation it comes up empty. So cannot be linked.
```
bwplotka, line 395:
```suggestion
	// If header is just punctuation it comes up empty, so it cannot be linked.
```

bwplotka (approved):
Thanks! LGTM

PR #53, opened 2021-06-24
Add profile.

bwplotka:


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/antchfx/xmlquery v1.3.4 // indirect
 	github.com/efficientgo/tools/core v0.0.0-20210609125236-d73259166f20
 	github.com/efficientgo/tools/extkingpin v0.0.0-20210609125236-d73259166f20
+	github.com/felixge/fgprof v0.9.1 // indirect
```
saswatamcode, line 10: Will need to run `go mod tidy` as linter fails here. 🙂

main.go
```diff
@@ -102,6 +115,23 @@ func main() {
 	}
 }
 
+func snapshotProfiles(dir string) (func() error, error) {
+	// TODO: now -> date
+	if err := os.MkdirAll(filepath.Join(dir, "now"), os.ModePerm); err != nil {
+		return nil, err
+	}
+	f, err := os.OpenFile(filepath.Join(dir, "now", "fgprof.pprof"), os.O_TRUNC|os.O_WRONLY, os.ModePerm)
```
saswatamcode, line 123:
```suggestion
	f, err := os.OpenFile(filepath.Join(dir, "now", "fgprof.pprof"), os.O_CREATE|os.O_TRUNC|os.O_WRONLY, os.ModePerm)
```
This would create a new file if none exists, which is why there was error before. 🙂

saswatamcode (approved):
Looks amazing! 💫  Just some suggestions.

main.go
saswatamcode: Screenshot 2021-06-25 at 10 49 36 AM
Works with Thanos now! 🙂

main.go
```diff
@@ -66,14 +78,28 @@ func main() {
 		Default("info").Enum("error", "warn", "info", "debug")
 	logFormat := app.Flag("log.format", "Log format to use.").
 		Default(logFormatCLILog).Enum(logFormatLogfmt, logFormatJson, logFormatCLILog)
+	// Profiling and metrics.
+	profilesPath := app.Flag("debug.profiles", "Path to which CPU and heap profiles are saved").Hidden().String()
+	metrics := app.Flag("metrics", "Enable metrics and view them at https://localhost:9091/metrics").Hidden().Bool()
```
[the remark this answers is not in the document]
bwplotka: Right I think we have to print it

main.go
```diff
@@ -101,6 +140,35 @@ func main() {
 		level.Error(logger).Log("err", errors.Wrapf(err, "%s command failed", cmd))
 		os.Exit(1)
 	}
+
+}
+
+func snapshotProfiles(dir string) (func() error, error) {
+	// TODO: now -> date
```
bwplotka, line 147: ?

pkg/mdformatter/linktransformer/link.go
```diff
@@ -286,6 +360,9 @@ func (v *validator) visit(filepath string, dest string, lineNumbers string) {
 	v.destFutures[k] = &futureResult{cases: 1, resultFn: func() error { return nil }}
 	matches := remoteLinkPrefixRe.FindAllStringIndex(dest, 1)
 	if matches == nil {
+		if v.l != nil {
```
bwplotka, line 363: What about doing it always and just never register/print metrics? (it will be much less distracting)

bwplotka (commented):
Nice, some suggestions only. Thanks for helping here!

main.go
```diff
@@ -34,6 +42,10 @@ const (
 	logFormatCLILog = "clilog"
 )
 
+type mdoxMetrics struct {
```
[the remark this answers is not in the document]
saswatamcode: I was kind of keeping the struct there since I was facing issues using just `reg` variable(metrics not registering due to how CLI operates). Will try out something else and see if I can remove.

main.go
saswatamcode: Missed this, will add it in. 🙂

pkg/mdformatter/linktransformer/link.go
saswatamcode: So no `--metrics` flag and it would always be enabled then?

main.go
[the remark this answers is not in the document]
saswatamcode: I've kept it for the `Print()` function now. 🙂

main.go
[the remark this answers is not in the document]
saswatamcode: Done!

main.go
saswatamcode: Done!

pkg/mdformatter/linktransformer/link.go
saswatamcode: Now it does this always but doesn't register if flag not provided! 🙂

main.go
```diff
@@ -101,6 +142,51 @@ func main() {
...
+func (m *mdoxMetrics) Print() error {
+	mfs, err := m.reg.Gather()
+	if err != nil {
+		return err
+	}
+
+	enc := expfmt.NewEncoder(os.Stdout, expfmt.FmtProtoText)
+
+	for _, mf := range mfs {
+		if err := enc.Encode(mf); err != nil {
```
bwplotka, line 185: We might want to make sure time is encoded too. By default there is no time encoded but in order to walk through those files at some point and load into block, we want timestamps

main.go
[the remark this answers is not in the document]
bwplotka: What about something simple: Always put that to the file,?

main.go
saswatamcode: Yes, I've thought of a few ways of adding timestamps! The server is removed and metrics are written to a file now. Will try to implement the timestamp mechanism!

pkg/mdformatter/mdformatter.go
```diff
@@ -257,6 +281,9 @@ func format(ctx context.Context, logger log.Logger, files []string, diffs *Diffs
 			spin.Message(fn + "...")
 		}
 		errs.Add(func() error {
+			start_time := time.Now()
```
bwplotka, line 284:
```suggestion
			startTime := time.Now()
```

pkg/mdformatter/mdformatter.go
```diff
@@ -288,6 +315,9 @@ func format(ctx context.Context, logger log.Logger, files []string, diffs *Diffs
 			if err != nil {
 				return errors.Wrapf(err, "write %v", fn)
 			}
+			time_taken := time.Since(start_time)
```
bwplotka, line 318: ditto

pkg/mdformatter/mdformatter.go
```diff
@@ -216,32 +239,33 @@ func newSpinner(suffix string) (*yacspin.Spinner, error) {
...
 		return err
 	}
-	return format(ctx, logger, files, nil, spin, opts...)
+	return format(ctx, logger, files, nil, spin, reg, opts...)
 }
 
 // IsFormatted tries to formats given markdown files and return Diff if files are not formatted.
 // If diff is empty it means all files are formatted.
-func IsFormatted(ctx context.Context, logger log.Logger, files []string, opts ...Option) (diffs Diffs, err error) {
+func IsFormatted(ctx context.Context, logger log.Logger, files []string, reg *prometheus.Registry, opts ...Option) (diffs Diffs, err error) {
```
bwplotka, line 252: Let's use option - metrics are only used rarely potentially (not by default)

bwplotka (commented):
LGTM, just minor nits! 💪🏽

main.go
```diff
@@ -66,14 +79,33 @@ func main() {
 		Default("info").Enum("error", "warn", "info", "debug")
 	logFormat := app.Flag("log.format", "Log format to use.").
 		Default(logFormatCLILog).Enum(logFormatLogfmt, logFormatJson, logFormatCLILog)
+	// Profiling and metrics.
+	profilesPath := app.Flag("debug.profiles", "Path to which CPU and heap profiles are saved").Hidden().String()
```
bwplotka, line 83:
```suggestion
	profilesPath := app.Flag("profiles.path", "Path to directory where CPU and heap profiles will be saved; If empty, no profiling will be enabled. ").ExistingDir()
```

main.go
```diff
@@ -66,14 +79,33 @@ func main() {
...
+	metrics := app.Flag("metrics", "Path to which metrics are saved in OpenMetrics format").Hidden().String()
```
bwplotka, line 84: Ditto for above changes (:

main.go
```diff
@@ -101,6 +133,65 @@ func main() {
...
+
+	fgFunc := fgprof.Start(f, fgprof.FormatPprof)
+
+	return func() (err error) {
+		defer errcapture.Do(&err, f.Close, "close")
+		return fgFunc()
+	}, nil
+}
+
+func (m *mdoxMetrics) Print() error {
```
bwplotka, line 167: Are we printing here? (::

bwplotka: I don't think you need struct.. just `Dump(r prometheus.Registry, dir string) error` would be more cleaner, WDYT?

main.go
```diff
@@ -101,6 +133,65 @@ func main() {
...
+	}
+	f, err := os.OpenFile(filepath.Join(m.dir, strings.ReplaceAll(now.Format(time.UnixDate), " ", "_"), "metrics"), os.O_CREATE|os.O_TRUNC|os.O_WRONLY, os.ModePerm)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	for _, mf := range mfs {
+		for _, metric := range mf.Metric {
+			unixTime := now.Unix()
```
bwplotka, line 184: Can we cache now, so we have consistent timestamp across all series?

main.go
```diff
@@ -101,6 +133,65 @@ func main() {
...
+			metric.TimestampMs = &unixTime
+		}
+		if _, err := expfmt.MetricFamilyToOpenMetrics(f, mf); err != nil {
+			return err
+		}
+	}
+	if _, err = expfmt.FinalizeOpenMetrics(f); err != nil {
```
bwplotka, line 191: For future: Let's document how one can use those and import to Prometheus 🤗

bwplotka (commented):
Let's go! Amazing, work and good example for everyone else. Let's write about it / blog / talk 🤩

main.go
saswatamcode: I think we do this already! `now` is defined when constructing dir name and is used across all metrics. So it's the same across all series.

main.go
saswatamcode: Done!

main.go
saswatamcode: Done!

main.go
saswatamcode: Done!

main.go
saswatamcode: Sure!

main.go
bwplotka: all good!

bwplotka (commented):
Nice! Let's go 💪🏽

PR #54, opened 2021-06-27
Add link caching using SQLite

saswatamcode:
This PR adds link caching using a local SQLite file, `.mdoxcache` which is suitable for GitHub Actions. It introduces the following:

Adds following fields to `links.validate.config`:
- `cacheType`:  Specifies type of cache. Can be set to 'SQLite' or 'None'. 'SQLite' creates a database in `.mdoxcache`. 'None' or setting it to empty will ensure no `.mdoxcache` file is created. Will throw error if set to anything else.
- `cacheValidity` which specifies duration till which link in the cache database is valid. Default is `120h` or 5 days.

Also adds following flag in case cache needs to be cleared but there is no filesystem access:
- `--cache.clear` which instructs mdox to drop the entire cache database and start afresh.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ vendor/
 
 .idea/
 .envrc
+mdoxcache
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,4 @@ vendor/
 
 .idea/
 .envrc
-mdoxcache
+.mdoxcache
--- a/.mdox.validate.yaml
+++ b/.mdox.validate.yaml
@@ -1,4 +1,5 @@
 version: 1
+noCache: true
 
 validators:
   # Validators to skip checking PR/issue links of mdox.
--- a/.mdox.validate.yaml
+++ b/.mdox.validate.yaml
@@ -1,5 +1,4 @@
 version: 1
-noCache: true
 
 validators:
   # Validators to skip checking PR/issue links of mdox.
--- a/README.md
+++ b/README.md
@@ -82,7 +82,10 @@ Flags:
                                  flag (mutually exclusive). Content of YAML file
                                  for skipping link check, with spec defined in
                                  github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig
-      --cache.clear              If true, entire cache database will be dropped
+      --cache.clear              If true, entire cache database will be dropped.
+                                 Useful in case mdox cache needs to be cleared
+                                 immediately from GitHub Actions or other CI
+                                 runner cache.
 
 Args:
   <files>  Markdown file(s) to process.
--- a/README.md
+++ b/README.md
@@ -82,10 +82,10 @@ Flags:
                                  flag (mutually exclusive). Content of YAML file
                                  for skipping link check, with spec defined in
                                  github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig
-      --cache.clear              If true, entire cache database will be dropped.
-                                 Useful in case mdox cache needs to be cleared
-                                 immediately from GitHub Actions or other CI
-                                 runner cache.
+      --cache.clear              If true, entire cache database will be dropped
+                                 and rebuilt when mdox is run. Useful in case
+                                 cache needs to be cleared immediately from
+                                 GitHub Actions or other CI runner cache.
 
 Args:
   <files>  Markdown file(s) to process.
--- a/README.md
+++ b/README.md
@@ -82,10 +82,6 @@ Flags:
                                  flag (mutually exclusive). Content of YAML file
                                  for skipping link check, with spec defined in
                                  github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig
-      --no-cache                 If true, link checking will not create local
-                                 SQLite cache file
-      --cache.validity=120h      Duration till which link in cache is considered
-                                 valid
       --cache.clear              If true, entire cache database will be dropped
 
 Args:
--- a/README.md
+++ b/README.md
@@ -82,6 +82,11 @@ Flags:
                                  flag (mutually exclusive). Content of YAML file
                                  for skipping link check, with spec defined in
                                  github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig
+      --no-cache                 If true, link checking will not create local
+                                 SQLite cache file
+      --cache.validity=120h      Duration till which link in cache is considered
+                                 valid
+      --cache.clear              If true, entire cache database will be dropped
 
 Args:
   <files>  Markdown file(s) to process.
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/bwplotka/mdox/pkg/cache"
 	"github.com/bwplotka/mdox/pkg/clilog"
 	"github.com/bwplotka/mdox/pkg/extkingpin"
 	"github.com/bwplotka/mdox/pkg/mdformatter"
@@ -40,6 +41,8 @@ const (
 	logFormatLogfmt = "logfmt"
 	logFormatJson   = "json"
 	logFormatCLILog = "clilog"
+
+	cacheFile = "mdoxcache"
 )
 
 func setupLogger(logLevel, logFormat string) log.Logger {
@@ -207,9 +210,13 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	anchorDir := cmd.Flag("anchor-dir", "Anchor directory for all transformers. PWD is used if flag is not specified.").ExistingDir()
 	linksLocalizeForAddress := cmd.Flag("links.localize.address-regex", "If specified, all HTTP(s) links that target a domain and path matching given regexp will be transformed to relative to anchor dir path (if exists)."+
 		"Absolute path links will be converted to relative links to anchor dir as well.").Regexp()
-	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig", extflag.WithEnvSubstitution())
+
+	cacheDisabled := cmd.Flag("no-cache", "If true, link checking will not create local SQLite cache file").Bool()
+	// Default validity is for 5 days.
+	cacheValidity := cmd.Flag("cache.validity", "Duration till which link in cache is considered valid").Default("120h").Duration()
+	clearCache := cmd.Flag("cache.clear", "If true, entire cache database will be dropped").Bool()
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var reg *prometheus.Registry
@@ -242,11 +249,23 @@ This directive runs executable with arguments and put its stderr and stdout outp
 
 		var linkTr []mdformatter.LinkTransformer
 		if *linksValidateEnabled {
+			var storage *cache.Storage
+
 			validateConfigContent, err := linksValidateConfig.Content()
 			if err != nil {
 				return err
 			}
-			v, err := linktransformer.NewValidator(ctx, logger, validateConfigContent, anchorDir, reg)
+
+			storage = nil
+			if !*cacheDisabled {
+				storage = &cache.Storage{
+					Filename:   cacheFile,
+					Validity:   *cacheValidity,
+					ClearCache: *clearCache,
+				}
+			}
+
+			v, err := linktransformer.NewValidator(ctx, logger, validateConfigContent, anchorDir, storage, reg)
 			if err != nil {
 				return err
 			}
--- a/main.go
+++ b/main.go
@@ -213,9 +213,6 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig", extflag.WithEnvSubstitution())
 
-	cacheDisabled := cmd.Flag("no-cache", "If true, link checking will not create local SQLite cache file").Bool()
-	// Default validity is for 5 days.
-	cacheValidity := cmd.Flag("cache.validity", "Duration till which link in cache is considered valid").Default("120h").Duration()
 	clearCache := cmd.Flag("cache.clear", "If true, entire cache database will be dropped").Bool()
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
@@ -256,13 +253,9 @@ This directive runs executable with arguments and put its stderr and stdout outp
 				return err
 			}
 
-			storage = nil
-			if !*cacheDisabled {
-				storage = &cache.Storage{
-					Filename:   cacheFile,
-					Validity:   *cacheValidity,
-					ClearCache: *clearCache,
-				}
+			storage = &cache.Storage{
+				Filename:   cacheFile,
+				ClearCache: *clearCache,
 			}
 
 			v, err := linktransformer.NewValidator(ctx, logger, validateConfigContent, anchorDir, storage, reg)
--- a/main.go
+++ b/main.go
@@ -246,14 +246,14 @@ This directive runs executable with arguments and put its stderr and stdout outp
 
 		var linkTr []mdformatter.LinkTransformer
 		if *linksValidateEnabled {
-			var storage *cache.Storage
+			var storage *cache.SQLite3Storage
 
 			validateConfigContent, err := linksValidateConfig.Content()
 			if err != nil {
 				return err
 			}
 
-			storage = &cache.Storage{
+			storage = &cache.SQLite3Storage{
 				Filename:   cacheFile,
 				ClearCache: *clearCache,
 			}
--- a/main.go
+++ b/main.go
@@ -42,7 +42,7 @@ const (
 	logFormatJson   = "json"
 	logFormatCLILog = "clilog"
 
-	cacheFile = "mdoxcache"
+	cacheFile = ".mdoxcache"
 )
 
 func setupLogger(logLevel, logFormat string) log.Logger {
@@ -213,7 +213,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig", extflag.WithEnvSubstitution())
 
-	clearCache := cmd.Flag("cache.clear", "If true, entire cache database will be dropped").Bool()
+	clearCache := cmd.Flag("cache.clear", "If true, entire cache database will be dropped. Useful in case mdox cache needs to be cleared immediately from GitHub Actions or other CI runner cache.").Bool()
 
 	cmd.Run(func(ctx context.Context, logger log.Logger) (err error) {
 		var reg *prometheus.Registry
--- a/main.go
+++ b/main.go
@@ -213,7 +213,7 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig", extflag.WithEnvSubstitution())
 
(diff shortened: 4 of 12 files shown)


pkg/cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+// Close cache database.
+func (s *Storage) Close() error {
+	err := s.dbHandle.Close()
+	return err
+}
+
+// Visited inserts new URL to cache database.
+func (s *Storage) Visited(requestID uint64) error {
+	// If particular URL is already inserted, then delete.
+	// Visit method will only be called if validity expires for a URL or in case of a new URL.
```
bwplotka, line 99: I would not assume the usage. Focus on Storage API. What would you expect from `Visited`. I would only expect... No idea, `visited` is the wrong name even as it suggests a bool return if something was visited. If that means just to `visit` something then we should probably bump the timestamp, nothing else, no?

pkg/cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+		return false, err
+	}
+
+	// Check if URL is within validity threshold.
+	now := time.Now().UTC()
+	if !timestamp.IsZero() && now.Sub(timestamp) <= s.Validity {
+		return true, nil
+	}
+	return false, nil
+}
```
bwplotka, line 140:
```suggestion
	return (!timestamp.IsZero() && ime.Now().UTC().Sub(timestamp) <= s.Validity), nil
}
```

pkg/mdformatter/linktransformer/link.go
```diff
@@ -180,6 +190,20 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 	v.c.OnError(func(response *colly.Response, err error) {
 		v.rMu.Lock()
 		defer v.rMu.Unlock()
+		// Colly stores requests in cache, even when they have errors.
+		// So need to delete requests with errors from cache database.
```
bwplotka, line 194: OMG this is quite annoying, can we add an issue on Colly to propose a different API?

pkg/mdformatter/linktransformer/link.go
```diff
@@ -180,6 +190,20 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
...
+		if v.storage != nil {
+			h := fnv.New64a()
```
bwplotka, line 196: Let's add that to separate method, ideally on storage side? (hashing request).

Are you sure it's the same as colly hash. What if they change that next version (quite prone to errro code)

pkg/mdformatter/linktransformer/link.go
```diff
@@ -180,6 +190,20 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
...
+			_, writeErr := h.Write([]byte(response.Ctx.Get(originalURLKey)))
+			if writeErr != nil {
+				v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(writeErr, "%q could not build requestID %v", response.Request.URL.String(), response.StatusCode)
+			}
+			delErr := v.storage.DeleteRequest(h.Sum64())
```
bwplotka, line 201: I wonder if it makes sense to use their interface if we have to do all those things manually.. Wouldn't be better to enhance our storage with API like error and store the status code etc too? Knowing fast that something is 404 is also coachable, no? (with maybe different validity)

bwplotka (commented):
Nice, great job! Commented some ideas.

In terms of testing, I would test things with and without such cache and use SQL lite file directly, any blocker for that?

README.md
```diff
@@ -69,6 +69,11 @@ Flags:
                                  flag (mutually exclusive). Content of YAML file
                                  for skipping link check, with spec defined in
                                  github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig
+      --no-cache                 If true, link checking will not create local
```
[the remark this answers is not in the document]
saswatamcode: Sure this can be done! 🙂

main.go
```diff
@@ -125,10 +128,14 @@ This directive runs executable with arguments and put its stderr and stdout outp
...
 	linksLocalizeForAddress := cmd.Flag("links.localize.address-regex", "If specified, all HTTP(s) links that target a domain and path matching given regexp will be transformed to relative to anchor dir path (if exists)."+
 		"Absolute path links will be converted to relative links to anchor dir as well.").Regexp()
-	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig", extflag.WithEnvSubstitution())
 
+	cacheDisabled := cmd.Flag("no-cache", "If true, link checking will not create local SQLite cache file").Bool()
+	// Default validity is for 5 days.
+	cacheValidity := cmd.Flag("cache.validity", "Duration till which link in cache is considered valid").Default("120h").Duration()
+	clearCache := cmd.Flag("cache.clear", "If true, entire cache database will be dropped").Bool()
```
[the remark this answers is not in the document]
saswatamcode: Yes locally, but I guess we cannot clear GitHub Actions cache yet since this is open: https://github.com/actions/cache/issues/2. So flag might be useful if we want to delete cache immediately from GH action and not wait till GitHub refreshes on its own. 🙂

pkg/cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+	_, err = statement.Exec()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Close cache database.
+func (s *Storage) Close() error {
+	err := s.dbHandle.Close()
```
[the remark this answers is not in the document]
saswatamcode: I guess we can remove!

pkg/cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+	return err
+}
+
+// Visited inserts new URL to cache database.
+func (s *Storage) Visited(requestID uint64) error {
```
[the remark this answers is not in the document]
saswatamcode: This is how colly does it (by hashing URLs for get requests). Seems weird to me too but cannot change as we need to conform to [Storage interface](https://github.com/gocolly/colly/blob/master/storage/storage.go#L30). They hash it [here](https://github.com/gocolly/colly/blob/master/colly.go#L735). 🙂

pkg/cache/cache.go
saswatamcode: Yes, `Visited` seems like the wrong name for this but I checked this [here](https://github.com/gocolly/colly/blob/master/colly.go#L755) and [here](https://github.com/velebak/colly-sqlite3-storage/blob/master/colly/sqlite3/sqlite3.go#L111), so seems like it actually inserts it into storage. So had to do it this way.

pkg/mdformatter/linktransformer/link.go
saswatamcode: Yes, I plan to add a few issues. 🙂

pkg/mdformatter/linktransformer/link.go
saswatamcode: I wanted to do it on storage side too, but have to conform colly storage interface so that we can pass that in.

pkg/mdformatter/linktransformer/link.go
saswatamcode: Yup. Was kind of thinking about the same thing, maybe would be better to just do this on our own and not pass any special storage to Colly. 🙂

pkg/cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+	// Create db with index.
+	statement, err := s.dbHandle.Prepare("CREATE TABLE IF NOT EXISTS visited (id INTEGER PRIMARY KEY, requestID INTEGER, visited INT, timestamp DATETIME)")
+	if err != nil {
+		return err
+	}
+	_, err = statement.Exec()
+	if err != nil {
+		return err
+	}
+	statement, err = s.dbHandle.Prepare("CREATE INDEX IF NOT EXISTS idx_visited ON visited (requestID)")
```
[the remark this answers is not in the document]
bwplotka: Done 
image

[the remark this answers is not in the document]
saswatamcode: Thank you!! ❤️

bwplotka:
```
#TODO(bwplotka): Uncomment once we upstream merrors package: @/tmp/.bin/faillint-v1.5.0 -paths "errors=github.com/pkg/errors" ./...
Error: /home/runner/work/mdox/mdox/pkg/mdformatter/linktransformer/link_test.go:225:50: too few arguments in call to MustNewValidator
faillint: error during loading
```
🤔 Hmmm, let me check today/Monday

bwplotka:
Let's try rebasing, maybe there is some old dep? (:

saswatamcode:
Yup! Rebasing fixed this!

pkg/cache/cache.go
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) Bartłomiej Płotka @bwplotka
+// Licensed under the Apache License 2.0.
+
+package cache
```
bwplotka, line 4: It would be amazing to get some feedback to their project (just mention, don't expect anything).

.mdox.validate.yaml
```diff
@@ -1,4 +1,5 @@
 version: 1
+noCache: true
```
bwplotka, line 2: Again, let's avoid boolean flags. What if we have another cache soon with different type? I believe there should be `cacheType` or something (: e.g `SQlite` and `None`

.gitignore
```diff
@@ -16,4 +16,5 @@ vendor/
 .bin
 
 .idea/
-.envrc
\ No newline at end of file
+.envrc
+mdoxcache
```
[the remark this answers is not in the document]
saswatamcode: Sure!

.mdox.validate.yaml
saswatamcode: Sure! Will implement this. 🙂

main.go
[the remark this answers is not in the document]
saswatamcode: Sure!

pkg/mdformatter/linktransformer/link.go
```diff
@@ -197,9 +204,13 @@ type futureResult struct {
 
 // NewValidator returns mdformatter.LinkTransformer that crawls all links.
 // TODO(bwplotka): Add optimization and debug modes - this is the main source of latency and pain.
-func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []byte, anchorDir string, reg *prometheus.Registry) (mdformatter.LinkTransformer, error) {
+func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []byte, anchorDir string, storage *cache.Storage, reg *prometheus.Registry) (mdformatter.LinkTransformer, error) {
 	var err error
-	config := Config{}
+	defaultValidity, err := time.ParseDuration("120h")
```
matej-g, line 209: Why not directly create a `const` with default `time.Duration`? It would save us from extra parsing here :blush:

pkg/mdformatter/linktransformer/link.go
```diff
@@ -239,6 +251,15 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 	if config.Timeout != "" {
 		v.c.SetRequestTimeout(config.timeout)
 	}
+
+	if v.validateConfig.CacheType == sqlite && storage != nil {
```
matej-g, line 255: Do we need the cache type check? :thinking: As long as `storage` is not `nil`, it should not matter what type we're initializing, right?

pkg/cache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	// Database handle.
+	dbHandle *sql.DB
+	// Mutex used for clearing cache database.
+	mu sync.RWMutex
+}
+
+// Init initializes cache database.
+func (s *Storage) Init() error {
+	// Check if db exists.
+	if s.dbHandle == nil {
```
matej-g, line 36: I'm wondering, if the database exists, do we need to go through the other steps below? Could we simply check if the database and table exists (optionally, truncate the table if `s.ClearCache` true) and if yes, exist, otherwise create new table? Would that flow make sense?

matej-g (commented):
Wonderful job as usual :hugs: I left some comments, would be cool to get it merged soon!

pkg/mdformatter/linktransformer/link.go
saswatamcode: Will do that!

pkg/cache/cache.go
saswatamcode: Let me try that out and get back! 🙂

pkg/mdformatter/linktransformer/link.go
matej-g: Hm then I would maybe expect `storage` to still be `nil`, after all if we do not want to initialize storage why would it not be `nil`? I think it's detail, it just feels like it could be simplified a bit, but it's more of a nit.

matej-g:
I'm wondering what's status of this PR? It was looking good enough for merging for me but I'm wondering if @bwplotka had a chance to look at the latest state as well? cc @saswatamcode

saswatamcode:
I didn't get a chance to look through the `dbHandle` flow that was mentioned in the last review. 😅
I think this is in a mergeable state once @bwplotka has had a look and we can iterate on this (will possibly need to once we use it in a repo and come across some new issues anyway), @matej-g ! 🙂

pkg/mdformatter/linktransformer/link.go
bwplotka: still not done?

pkg/mdformatter/linktransformer/link.go
bwplotka: still to be done?

pkg/mdformatter/linktransformer/link.go
```diff
@@ -239,6 +251,15 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
...
+		v.storage = storage
+		v.storage.Validity = v.validateConfig.CacheValidity
```
bwplotka, line 257: This looks like a code smell.

We pass storage here for some composability, but we kind of  use it like struct (setting Field directly) despite doing  `Init`  below. Can we add `Validity` parameter to `Init` then?

pkg/mdformatter/linktransformer/link.go
```diff
@@ -253,6 +274,12 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 	v.c.OnScraped(func(response *colly.Response) {
 		v.rMu.Lock()
 		defer v.rMu.Unlock()
+		if v.storage != nil {
+			err := v.storage.CacheURL(response.Ctx.Get(originalURLKey))
+			if err != nil {
```
bwplotka, line 279:
```suggestion
			if err := v.storage.CacheURL(response.Ctx.Get(originalURLKey)); err != nil {
```

pkg/cache/cache.go
```diff
@@ -0,0 +1,143 @@
...
+	row := statement.QueryRow(URL)
+	if err = row.Scan(&timestamp); err != nil {
+		// If ErrNoRows then it means URL is new, so need to call Visited.
+		if err == sql.ErrNoRows {
+			return false, nil
+		}
+		return false, err
+	}
+
+	// Check if URL is within validity threshold.
```
bwplotka, line 131: I would add some jitter to it (and jitter might be parameter). The reason is that we will have timeouts still if we will expire all items in one go.

bwplotka (approved):
Nice! IMO it's good to be merged, just minor things (: 

Thanks!

pkg/mdformatter/linktransformer/link.go
saswatamcode: Yes, needs to be done!

pkg/mdformatter/linktransformer/link.go
saswatamcode: We can iterate on this in separate PR!

main.go
```diff
@@ -207,10 +210,11 @@ This directive runs executable with arguments and put its stderr and stdout outp
 	anchorDir := cmd.Flag("anchor-dir", "Anchor directory for all transformers. PWD is used if flag is not specified.").ExistingDir()
 	linksLocalizeForAddress := cmd.Flag("links.localize.address-regex", "If specified, all HTTP(s) links that target a domain and path matching given regexp will be transformed to relative to anchor dir path (if exists)."+
 		"Absolute path links will be converted to relative links to anchor dir as well.").Regexp()
-	// TODO(bwplotka): Add cache in file?
 	linksValidateEnabled := cmd.Flag("links.validate", "If true, all links will be validated").Short('l').Bool()
 	linksValidateConfig := extflag.RegisterPathOrContent(cmd, "links.validate.config", "YAML file for skipping link check, with spec defined in github.com/bwplotka/mdox/pkg/linktransformer.ValidatorConfig", extflag.WithEnvSubstitution())
 
+	clearCache := cmd.Flag("cache.clear", "If true, entire cache database will be dropped. Useful in case mdox cache needs to be cleared immediately from GitHub Actions or other CI runner cache.").Bool()
```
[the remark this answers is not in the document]
saswatamcode: Will update!

pkg/mdformatter/linktransformer/link.go
saswatamcode: Will try to restructure this!

pkg/cache/cache.go
saswatamcode: Will do!

bwplotka:
Still to be addressed I believe (:

matej-g:
Any chance of progress? Before this PR celebrates it's 1st birthday in a week :birthday: :joy:

saswatamcode:
Oh no! Yup, let me get to it soon, it's been on my backlog for a long time now! 🙈

saswatamcode:
@bwplotka @matej-g I've addressed most of the feedback here. But yes, there are some code smells due to cache, that I think arise from not having a dedicated `Cache` interface. We try to support multiple `CacheType`s here but use only the sqlite `Storage` struct, which makes it a bit confusing. 
Let's get this merged, I will refactor this in another PR, without changing configuration semantics. 🙂

pkg/cache/cache.go
```diff
@@ -0,0 +1,150 @@
...
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/pkg/errors"
+)
+
+const (
+	driverName = "sqlite3"
+)
+
+// Storage implements a SQLite3 caching backend for Colly.
+type Storage struct {
```
bwplotka, line 20: Sqlite3Storage

pkg/cache/cache.go
```diff
@@ -0,0 +1,150 @@
...
+	row := statement.QueryRow(URL)
+	if err = row.Scan(&timestamp); err != nil {
+		// If ErrNoRows then it means URL is new, so need to call Visited.
+		if err == sql.ErrNoRows {
+			return false, nil
+		}
+		return false, err
+	}
+
+	// Check if URL is within validity threshold with jitter.
```
bwplotka, line 138: If we want jitter, let's randomize it from [0 -> Validity] (:

bwplotka (approved):
Amazing, thanks! Small nits only.

pkg/mdformatter/linktransformer/config.go
```diff
@@ -69,14 +77,19 @@ const (
 
 const (
 	gitHubAPIURL = "https://api.github.com/repos/%v/%v?sort=created&direction=desc&per_page=1"
+
+	none   = "None"
+	sqlite = "SQLite"
+
+	defaultValidity = time.Duration(432000000000000)
```
douglascamata, line 84: Better to convert this duration into something more readable: 
```suggestion
	defaultValidity = 2 * time.Hour
```

pkg/cache/cache.go
```diff
@@ -0,0 +1,158 @@
...
+	rand.Seed(time.Now().UnixNano())
+	if s.Jitter == time.Duration(0) {
+
+		// In case a custom jitter is not provided, a jitter from 0 to cache validity is used.
+		return (!timestamp.IsZero() && time.Now().UTC().Sub(timestamp)+time.Duration(rand.Intn(int(s.Validity))) <= s.Validity), nil
```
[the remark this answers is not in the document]
douglascamata: Another tip, `time.Now().UTC().Sub(timestamp)` is the same as `time.Since(timestamp)`.

pkg/mdformatter/linktransformer/config.go
douglascamata: Could even create a separate constant for the duration of a day (there's no `time.Day`) to make it better.

pkg/cache/cache.go
```diff
@@ -0,0 +1,158 @@
...
+	if err = row.Scan(&timestamp); err != nil {
+		// If ErrNoRows then it means URL is new, so need to call Visited.
+		if err == sql.ErrNoRows {
+			return false, nil
+		}
+		return false, err
+	}
+
+	// Check if URL is within validity threshold with jitter.
+	rand.Seed(time.Now().UnixNano())
```
bwplotka, line 140: This should be done in the init of struct and ideally on non-global rand if possible (:

pkg/cache/cache.go
bwplotka, line 144: Plus I think it might make sense to just have one comparison with `jitterValue` being 0 for jitter equal 0 and value rand[0-jitter] otherwise (:

bwplotka (approved):
Amazing.

+1 to @douglascamata comments, plus added few more. Thanks!

bwplotka (approved):
Tiny nit, not a blocker, LGTM! Thanks

PR #57, opened 2021-07-04
Add `addPath` option

### LINKED ISSUE #51: Add new file via transform config
Some website pre-processes might require an extra file that isn't exactly transformed from some other file but is newly added with some frontmatter(maybe to maintain order with `weight`).

saswatamcode:
This PR adds `addPath` option to `transform` config which allows adding a new file to `outputDir` with `frontMatter` using config like below,

```
  - addPath: _index.md
    frontMatter:
      template: |
        weight: 100000
```

It cannot be specified alongside `path` or `glob`.
Only `Origin.LastMod` and `target.FileName` are specifiable in `addPath` frontMatter template. Fixes #51.


pkg/transform/config.go
```diff
@@ -68,6 +68,10 @@ type TransformationConfig struct {
 	// TODO(bwplotka): Explain ** and * suffixes and ability to specify "invalid" paths like "/../".
 	Path string
 
+	// Add is an optional path for a file to be created with certain frontMatter.
+	// Only parses frontMatter and cannot be mentioned alongside path and glob.
+	AddPath string `yaml:"addPath"`
```
bwplotka, line 73: Thanks for this work, but I think I am not a fan of this approach. This is a little bit mix of functionalities. We do go through transformations and we don't transform any old things here. Also why the only frontmatter if the whole file is empty? Also where the new file is put? In provided path? What if we want to provide a file next to another file we traverse?

I think it will be unwise to abstract all Linux commands within the transform configuration ): It's so easy to create this file with simple `echo ... > file`.

One functionality which would be harder with Linux commands (harder than 3 words) is to add certain file content next to another file we traverse with glob 🤔 

I think I would prefer something simple for now like `ExtraFiles` with `content` template or something like this 🤔

bwplotka (commented):
Thanks for this work, but not sure completely about this.

pkg/transform/config.go
saswatamcode: > Also why the only frontmatter if the whole file is empty? 

Creating a new file, with just frontmatter, was the use case I was aiming to fulfill with this, as seen [here](https://github.com/thanos-io/thanos/blob/main/scripts/website/contentpreprocess.sh#L86) in Thanos.
> Also where the new file is put? In provided path? What if we want to provide a file next to another file we traverse?

Yes, it is put in the provided path. For putting a new file next to a particular file we traverse, I think we can just mention path of the file we want to place it next to...but yes, `addPath` cannot work with glob.

> I think it will be unwise to abstract all Linux commands within the transform configuration ): It's so easy to create this file with simple `echo ... > file`. 

Yes, I see your point. Will try to make this simpler with `ExtraFiles` and `content`! 🙂

saswatamcode:
Closing as discussed in 1:1. 🙂

PR #58, opened 2021-07-05
Add option to preserve header

saswatamcode:
This PR changes `transform` tranformation config to preserve markdown headers by default while transforming as discussed in 1:1. `popHeader` option may be added to enable removing header. 

Note: Could not find a better explicit way of enabling remove header other than bool option in config (as was mentioned in 1:1). Would a flag instead be preferred?

### DIFF
--- a/pkg/transform/testdata/mdox1.yaml
+++ b/pkg/transform/testdata/mdox1.yaml
@@ -8,7 +8,6 @@ gitIgnored: true
 transformations:
   - glob: "README.md"
     path: _index.md
-    popHeader: true
     frontMatter:
       template: |
         title: "{{ .Origin.FirstHeader }}"
--- a/pkg/transform/testdata/mdox1.yaml
+++ b/pkg/transform/testdata/mdox1.yaml
@@ -8,6 +8,7 @@ gitIgnored: true
 transformations:
   - glob: "README.md"
     path: _index.md
+    popHeader: true
     frontMatter:
       template: |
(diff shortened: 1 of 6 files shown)


bwplotka (approved):
Nice LGTM, just one idea: Wonder if we can global default `popHeader` on root file? 🤔

saswatamcode:
Now, `popHeader` will be true by default for root files in input dir. This can be disabled by setting `popHeader` to false with glob matching root file. Also, all root files will have this as true by default, so would need to be configured separately if this behavior is not needed. Test case added in `pkg/transform/testdata/mdox3.yaml` where `popHeader` is not set for `README.md` but md header is still popped.

pkg/transform/testdata/mdox1.yaml
```diff
@@ -8,6 +8,7 @@ gitIgnored: true
 transformations:
   - glob: "README.md"
     path: _index.md
+    popHeader: true
```
bwplotka, line 11: If popHeader is true on root, why we do this? 🤔

saswatamcode: Sorry, forgot to remove!

bwplotka: It's still there on all examples 🤔

bwplotka:
Merging, LGTM thanks, but examples are  still not perfect 🙈

saswatamcode:
I've missed some. Will raise PR to fix them! 🙂

PR #59, opened 2021-07-10
Make GitHub validator smarter

saswatamcode:
This PR makes the special `github` validator smarter. This renames `github` validator to `githubPullsIssues`. (As discussed in 1:1.)

Configuration will now look like,
```
- regex: '(^http[s]?:\/\/)(www\.)?(gitlab\.com\/)thanos-io\/thanos(\/pull\/|\/issues\/)'
    type: 'githubPullsIssues'
- regex: '(^http[s]?:\/\/)(www\.)?(github\.com\/)thanos-io\/thanos(\/releases\/tag\/)'
    type: 'ignore'
```
This also avoids surprises by asking users to provide the correct regex so that `mdox` does not have to create it.

In case of faulty regex, below errors are shown,

For `githubPullsIssues`,
```
error: fmt command failed: parsing githubPullIssues Regex: GitHub PR/Issue regex not valid. Correct regex: (^http[s]?:\/\/)(www\.)?(github\.com\/){ORG_NAME}\/{REPO_NAME}(\/pull\/|\/issues\/)
```

### DIFF
--- a/pkg/mdformatter/linktransformer/config.go
+++ b/pkg/mdformatter/linktransformer/config.go
@@ -23,14 +23,14 @@ type Config struct {
 }
 
 type ValidatorConfig struct {
-	// Regex for type github is reponame matcher, like `bwplotka\/mdox`.
+	// Regex for type of validator. For `githubPullsIssues` this is: (^http[s]?:\/\/)(www\.)?(github\.com\/){ORG_NAME}\/{REPO_NAME}(\/pull\/|\/issues\/).
 	Regex string `yaml:"regex"`
-	// By default type is `ignore`. Could be `github` or `roundtrip`.
+	// By default type is `roundtrip`. Could be `githubPullsIssues` or `ignore`.
 	Type ValidatorType `yaml:"type"`
 	// GitHub repo token to avoid getting rate limited.
 	Token string `yaml:"token"`
 
-	ghValidator GitHubValidator
+	ghValidator GitHubPullsIssuesValidator
 	rtValidator RoundTripValidator
 	igValidator IgnoreValidator
 }
@@ -39,7 +39,7 @@ type RoundTripValidator struct {
 	_regex *regexp.Regexp
 }
 
-type GitHubValidator struct {
+type GitHubPullsIssuesValidator struct {
 	_regex  *regexp.Regexp
 	_maxNum int
 }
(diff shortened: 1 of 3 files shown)


pkg/mdformatter/linktransformer/config.go
```diff
@@ -23,24 +23,30 @@ type Config struct {
...
-	// Regex for type github is reponame matcher, like `bwplotka\/mdox`.
+	// Regex for type of validator. For `githubPullsIssues` this is: (^http[s]?:\/\/)(www\.)?(github\.com\/){ORG_NAME}\/{REPO_NAME}(\/pull\/|\/issues\/).
 	Regex string `yaml:"regex"`
-	// By default type is `ignore`. Could be `github` or `roundtrip`.
+	// By default type is `roundtrip`. Could be `githubPullsIssues` or `ignore`.
 	Type ValidatorType `yaml:"type"`
 	// GitHub repo token to avoid getting rate limited.
 	Token string `yaml:"token"`
+	// Validators to be used alongside githubPullsIssues.
+	NPIValidators []NonPullsIssuesValidator `yaml:"nonPullsIssuesValidators"`
```
bwplotka, line 33: hmmm, I am not fan of this approach...

Reasons:
* We are leaking GH special flow. It's actually quite surprise to guess what is the logic here
* If you regex only fo pulls or issues how other links even get there? In your example I would expect only pulls and issues being matched here so nonXYZ thing will never get called (at least that's my perception)

```
  - regex: '(^http[s]?:\/\/)(www\.)?(gitlab\.com\/)thanos-io\/thanos(\/pull\/|\/issues\/)'
    type: 'githubPullsIssues'
    nonPullsIssuesValidators:
      - regex: '(^http[s]?:\/\/)(www\.)?(github\.com\/)thanos-io\/thanos(\/releases\/tag\/)'
      - regex: '(^http[s]?:\/\/)(www\.)?(github\.com\/)thanos-io\/thanos(\/blob\/)'
      - regex: '(^http[s]?:\/\/)(www\.)?(github\.com\/)thanos-io\/thanos(\/commit\/)'```
```

Do we need this. Why not simple approach of regexing pull and issues only (and erroring out when user passes non issue or non PR) then use ignore type to ignore further stuff? 🤔

bwplotka (changes requested):
I think it's .... too smart. Too smart for me 🙈 

Some comments (: I think we talked about something different in 1:1.

But good direction! (:

pkg/mdformatter/linktransformer/config.go
saswatamcode: Yes, I was kind of mixing nested validators + `pullsIssues` which on second thought might make things super confusing 😅. `ignore` would be sufficient for the `nonPullsIssues` thing. Will remove this and just have `githubPullsIssues`. So example config would be,
```
  - regex: '(^http[s]?:\/\/)(www\.)?(gitlab\.com\/)thanos-io\/thanos(\/pull\/|\/issues\/)'
    type: 'githubPullsIssues'
  - regex: '(^http[s]?:\/\/)(www\.)?(github\.com\/)thanos-io\/thanos(\/releases\/tag\/)'
    type: 'ignore'
```

bwplotka (approved):
Nice, let's go!

PR #60, opened 2021-07-12
Generate YAML from Go struct

### LINKED ISSUE #23: Add subcommand that will generate example YAML that will parsable from Go structs

saswatamcode:
This PR allows mdox to generate YAML from Go structs, using the following semantics(for now),
```
```yaml mdox-gen-go-struct="main.go:Config"
```
Dependencies added:[jennifer](https://github.com/dave/jennifer),  [structtag](https://github.com/fatih/structtag)

Few TODOs:
```
// TODO(saswatamcode): Add tests.
// TODO(saswatamcode): Check jennifer code for some safety.
// TODO(saswatamcode): Add mechanism for caching output from generated code.
// TODO(saswatamcode): Currently takes file names, need to make it module based(something such as https://golang.org/pkg/cmd/go/internal/list/).
// TODO(saswatamcode): Better errors.
```

Resolves #23.

### DIFF
--- a/pkg/yamlgen/yamlgen.go
+++ b/pkg/yamlgen/yamlgen.go
@@ -1,0 +1,221 @@
+// Copyright (c) Bartłomiej Płotka @bwplotka
+// Licensed under the Apache License 2.0.
+
+package yamlgen
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/dave/jennifer/jen"
+	"github.com/pkg/errors"
+)
+
+// TODO(saswatamcode): Add tests.
+// TODO(saswatamcode): Check jennifer code for some safety.
+// TODO(saswatamcode): Add mechanism for caching output from generated code.
+// TODO(saswatamcode): Currently takes file names, need to make it module based(something such as https://golang.org/pkg/cmd/go/internal/list/).
+
+// GenGoCode generates Go code for yaml gen from structs in src file.
+func GenGoCode(src []byte) (string, error) {
+	// Create new main file.
+	fset := token.NewFileSet()
+	generatedCode := jen.NewFile("main")
+
+	// Parse source file.
+	f, err := parser.ParseFile(fset, "", src, parser.AllErrors)
+	if err != nil {
+		return "", err
+	}
+
+	// Add imports if needed(will not be used if not required in code).
+	for _, s := range f.Imports {
+		generatedCode.ImportName(s.Path.Value[1:len(s.Path.Value)-1], "")
+	}
+
+	// Init statements for structs.
+	var init []jen.Code
+	// Declare config map, i.e, `configs := map[string]interface{}{}`.
+	init = append(init, jen.Id("configs").Op(":=").Map(jen.String()).Interface().Values())
+
+	// Loop through declarations in file.
+	for _, decl := range f.Decls {
+		// Cast to generic declaration node.
+		if genericDecl, ok := decl.(*ast.GenDecl); ok {
+			// Check if declaration spec is `type`.
+			if typeDecl, ok := genericDecl.Specs[0].(*ast.TypeSpec); ok {
+				var structFields []jen.Code
+				// Cast to `type struct`.
+				structDecl := typeDecl.Type.(*ast.StructType)
+				fields := structDecl.Fields.List
+				// Loop and generate fields for each field.
+				for _, field := range fields {
+					// Each field might have multiple names.
+					names := field.Names
+					for _, n := range names {
+						pos := n.Obj.Decl.(*ast.Field)
+						structFields = append(structFields, jen.Id(n.Name).Id(string(src[pos.Type.Pos()-1:pos.Type.End()-1])).Id(string(src[pos.Tag.Pos()-1:pos.Tag.End()-1])))
+					}
+				}
+
+				// Add initialize statements for struct.
+				init = append(init, jen.Id("configs").Index(jen.Lit(typeDecl.Name.Name)).Op("=").Id(typeDecl.Name.Name+"{}"))
+				// Finally put struct inside generated code.
+				generatedCode.Type().Id(typeDecl.Name.Name).Struct(structFields...)
+			}
+		}
+	}
+
+	// Add for loop to iterate through map and return config YAML.
+	init = append(init, jen.For(
+		jen.List(jen.Id("k"), jen.Id("config")).Op(":=").Range().Id("configs"),
+	).Block(
+		// We import the cfggen Generate method directly to generate output.
+		jen.Qual("fmt", "Println").Call(jen.Lit("---")),
+		jen.Qual("fmt", "Println").Call(jen.Id("k")),
+		// TODO(saswatamcode): Replace with import from mdox itself once merged.
+		// jen.Qual("github.com/bwplotka/mdox/yamlgen", "Generate").Call(jen.Id("config"), jen.Qual("os", "Stderr")),
+		jen.Qual("structgen/cfggen", "Generate").Call(jen.Id("config"), jen.Qual("os", "Stderr")),
+	))
+
+	// Generate main function in new module.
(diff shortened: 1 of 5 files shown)


pkg/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/pkg/errors"
+)
+
+// TODO(saswatamcode): Add tests.
+// TODO(saswatamcode): Check jennifer code for some safety.
+// TODO(saswatamcode): Add mechanism for caching output from generated code.
+// TODO(saswatamcode): Currently takes file names, need to make it module based(something such as https://golang.org/pkg/cmd/go/internal/list/).
+
+// GenGoCode generates Go code for yaml gen from structs in src file.
+func GenGoCode(src []byte) (string, error) {
```
bwplotka, line 30: Good for start, but I wonder if we can scope to just one desired structure, to limit imports. Maybe it's later optimization - so fine for now

pkg/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,243 @@
...
+		generatedCode.ImportName(s.Path.Value[1:len(s.Path.Value)-1], "")
+	}
+
+	// Init statements for structs.
+	var init []jen.Code
+	// Declare config map, i.e, `configs := map[string]interface{}{}`.
+	init = append(init, jen.Id("configs").Op(":=").Map(jen.String()).Interface().Values())
+
+	// Loop through declarations in file.
+	for _, decl := range f.Decls {
```
bwplotka, line 52: I think I would expect some recursion here (:

pkg/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,243 @@
...
+		jen.Qual("fmt", "Println").Call(jen.Id("k")),
+		// TODO(saswatamcode): Replace with import from mdox itself once merged.
+		// jen.Qual("github.com/bwplotka/mdox/yamlgen", "Generate").Call(jen.Id("config"), jen.Qual("os", "Stderr")),
+		jen.Qual("structgen/cfggen", "Generate").Call(jen.Id("config"), jen.Qual("os", "Stderr")),
+	))
+
+	// Generate main function in new module.
+	generatedCode.Func().Id("main").Params().Block(
+		init...,
+	)
```
bwplotka, line 114:
```suggestion
	generatedCode.Func().Id("main").Params().Block(init...)
```

pkg/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,243 @@
...
+	cmd.Dir = tmpDir
+	if err := cmd.Run(); err != nil {
+		return nil, errors.Wrapf(err, "run %v", cmd)
+	}
+
+	// Import required packages(generate go.sum).
+	cmd = exec.CommandContext(ctx, "go", "mod", "tidy")
+	cmd.Dir = tmpDir
+	if err := cmd.Run(); err != nil {
+		return nil, errors.Wrapf(err, "run %v", cmd)
```
bwplotka, line 166:
```suggestion
		return nil, errors.Wrapf(err, "mod %v", cmd)
```

pkg/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,243 @@
...
+	_, err = main.Write([]byte(mainGo))
+	if err != nil {
+		return nil, err
+	}
+
+	// Create go.mod in temp dir.
+	cmd := exec.CommandContext(ctx, "go", "mod", "init", "structgen")
+	cmd.Dir = tmpDir
+	if err := cmd.Run(); err != nil {
+		return nil, errors.Wrapf(err, "run %v", cmd)
```
bwplotka, line 159:
```suggestion
		return nil, errors.Wrapf(err, "mod init %v", cmd)
```

pkg/yamlgen/yamlgen.go
```diff
@@ -0,0 +1,243 @@
...
+		return errors.Wrap(err, "invalid type")
+	}
+	return yaml.NewEncoder(w).Encode(obj)
+}
+
+func checkForOmitEmptyTagOption(obj interface{}) error {
+	return checkForOmitEmptyTagOptionRec(reflect.ValueOf(obj))
+}
+
+func checkForOmitEmptyTagOptionRec(v reflect.Value) error {
```
bwplotka, line 209: Do we need to check for emit empty if we fill all the structs? I think in this case we need to also set non default values for other things that slices... so maybe for next iteration

bwplotka (commented):
That's some next level magic

AI programmer!

LGTM, just let's add tests 👍🏽

pkg/yamlgen/yamlgen.go
saswatamcode: Done!

pkg/yamlgen/yamlgen.go
saswatamcode: Done!

pkg/yamlgen/yamlgen.go
saswatamcode: Done!

pkg/yamlgen/yamlgen.go
bwplotka: Can we add todo for that?

pkg/yamlgen/yamlgen_test.go
```diff
@@ -0,0 +1,60 @@
...
+		source := []byte("package main\n\ntype TestConfig struct {\n\tUrl   string `yaml:\"url\"`\n\tID    int    `yaml:\"id\"`\n\tToken string `yaml:\"token\"`\n}\n")
+		generatedCode, err := GenGoCode(source)
+		testutil.Ok(t, err)
+
+		expected := "package main\n\nimport (\n\t\"fmt\"\n\tyamlgen \"github.com/bwplotka/mdox/pkg/yamlgen\"\n\t\"os\"\n)\n\ntype TestConfig struct {\n\tUrl   string `yaml:\"url\"`\n\tID    int    `yaml:\"id\"`\n\tToken string `yaml:\"token\"`\n}\n\nfunc main() {\n\tconfigs := map[string]interface{}{}\n\tconfigs[\"TestConfig\"] = TestConfig{}\n\tfor k, config := range configs {\n\t\tfmt.Println(\"---\")\n\t\tfmt.Println(k)\n\t\tyamlgen.Generate(config, os.Stderr)\n\t}\n}\n"
+		testutil.Equals(t, expected, generatedCode)
+	})
+
+	t.Run("struct with unexported field", func(t *testing.T) {
+		source := []byte("package main\n\nimport \"regexp\"\n\ntype ValidatorConfig struct {\n\tType  string `yaml:\"type\"`\n\tRegex string `yaml:\"regex\"`\n\tToken string `yaml:\"token\"`\n\n\tr *regexp.Regexp\n}\n")
```
bwplotka, line 24: I would use ` and add +  backtick + , something like that. Still easier than long string manually crafted.

Another idea - put all in files - that would do as well.

bwplotka (commented):
LGTM, let's add recursion and understand what map will work, what not. 👍🏽

Dentrax:
Any update on this? @bwplotka

bwplotka:
I think @saswatamcode has some comments to be addressed, but quite close.

What do you need this for @Dentrax? Curious about your use case and if this PR fixes it (:

Dentrax:
I've created a related issue https://github.com/thanos-io/thanos/issues/4751#issuecomment-967799915 and closed due to inactivity. We want to add support for both comments and default values while generating YAML.

Dentrax:
Kind ping, still looking for this feature! @saswatamcode @bwplotka

bwplotka:
cc @saswatamcode

saswatamcode:
I'm trying to look into this, this week and bring it to a nice usable state, rebasing and correcting some of the old code. Plan to clean it up in the next few commits! 🙂

PR #63, opened 2021-07-23
Add concurrency to `mdformatter`

saswatamcode:
This PR parallelizes the file processing loop in `mdformatter` for faster formatting and link checking using `WaitGroup`. (This also changes CLI spinner behavior as filenames cannot be shown due to multiple files being processed at once)

Some caveats,
- Cancellation isn't immediate when link checking is enabled
- Panic during cancellation sometimes due to `v.c.Wait()`(flaky, maybe due to internal colly WaitGroup)
- Need some way to limit number of file processing goroutines


pkg/mdformatter/linktransformer/link.go
```diff
@@ -279,19 +290,27 @@ func (v *validator) visit(filepath string, dest string, lineNumbers string) {
...
-		v.destFutures[k].cases++
+	// If key present, delete and increment cases.
+	if prevResult, loaded := v.destFutures.LoadAndDelete(k); loaded {
+		newResult := prevResult.(*futureResult)
+		newResult.cases++
+		v.destFutures.Store(k, newResult)
 		return
 	}
-	v.destFutures[k] = &futureResult{cases: 1, resultFn: func() error { return nil }}
+
```
bwplotka, line 300: I think there is a race during replacement operation.

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -34,7 +34,11 @@ func (v GitHubValidator) IsValid(k futureKey, r *validator) (bool, error) {
 // RoundTripValidator.IsValid returns true if url is checked by colly.
 func (v RoundTripValidator) IsValid(k futureKey, r *validator) (bool, error) {
 	// Result will be in future.
-	r.destFutures[k].resultFn = func() error { return r.remoteLinks[k.dest] }
+	prevResult, _ := r.destFutures.LoadAndDelete(k)
```
bwplotka, line 37: Lot's of races when we delete, unfortunately.. we need to make this operation locked potentially with Mutex. 

But it has to be the full operation Locked:
1. Checking if there is entry in map
2. If no, create new future, and schedule colly
3. If yes, grab reference to future

We might need to check `IsValid` interface, if this actually allows us to do this 🤔

bwplotka (commented):
Looks good so far! Thanks, some comments.

bwplotka:
Still comments not addressed, I believe?

bwplotka:
Any progress?

PR #68, opened 2021-08-01
Add newline check for `mdox-exec` output

### LINKED ISSUE #67: Check newline in `mdox-exec` output
Some commands do not append a newline character by default, for example, `cat`, 
So a file like this which also doesn't have a new line at the end,
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: grafana
spec:
  ingress:
  - ports:
    - port: 3000
      protocol: TCP
  podSelector:
    matchLabels:
      app: grafana
```
is formatted like this
```
```yaml mdox-exec="cat test.yaml"
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: grafana
spec:
  ingress:
  - ports:
    - port: 3000
      protocol: TCP
  podSelector:
    matchLabels:
      app: grafana```
```
Might need to check output of command for newline.

saswatamcode:
This PR adds a check to ensure `mdgen` output for a command always has a new line at the end so that there are no surprising formatting issues. Test case added. Fixes #67.

### DIFF
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -84,7 +84,12 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 			}
 			return nil, errors.Wrapf(err, "run %v, out: %v", execCmd, b.String())
 		}
-		return b.Bytes(), nil
+		output := b.Bytes()
+		// Add newline to output if not present.
+		if !bytes.HasSuffix(output, []byte("\n")) {
+			output = append(output, []byte("\n")...)
+		}
+		return output, nil
 	}
 
 	panic("should never get here")
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -20,6 +20,10 @@ const (
 	infoStringKeyExitCode = "mdox-expect-exit-code"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -84,7 +84,12 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 			}
 			return nil, errors.Wrapf(err, "run %v, out: %v", execCmd, b.String())
 		}
-		return b.Bytes(), nil
+		output := b.Bytes()
+		// Add newline to output if not present.
+		if !bytes.HasSuffix(output, []byte("\n")) {
```
bwplotka, line 89:
```suggestion
		if !bytes.HasSuffix(output, newLineChar) {
```
.. to not creat slice every time? (micro-optimization, but also readability) and I think we already have newLineChar somewhere 🤔

bwplotka (approved):
LGTM!

pkg/mdformatter/mdgen/mdgen.go
bwplotka: not a blocker

saswatamcode: Done!

bwplotka (approved):
thanks!

PR #76, opened 2021-08-15
Add `linkPrefixForNonMarkdownResources` to transform config

### LINKED ISSUE #74: Support for non-doc local git links
https://github.com/thanos-io/thanos/pull/4544#discussion_r685919685

saswatamcode:
This PR adds `linkPrefixForNonMarkdownResources` option to mdox transform config. This allows `mdox` to prefix non-markdown or image relative links with URL. 
So with config set to`linkPrefixForNonMarkdownResources: https://github.com/bwplotka/mdox/tree/main`, relative links like `pkg/mdformatter/mdformatter.go` will be converted to `https://github.com/bwplotka/mdox/tree/main/pkg/mdformatter/mdformatter.go`. Fixes #74.


pkg/transform/testdata/mdox4.yaml
```diff
@@ -0,0 +1,51 @@
+version: 1
+
+inputDir: "testdata/testproj"
+outputDir: "testdata/tmp/test4/4"
+extraInputGlobs:
+  - "testdata/test.md"
+  - "testdata/teststatic"
+
+glueLink: "https://github.com/bwplotka/mdox/tree/main"
```
bwplotka, line 9: Let's rethink the name.. maybe `linkPrefixForNonMarkdownResources`

pkg/transform/transform.go
```diff
@@ -296,6 +307,30 @@ func (r *relLinkTransformer) TransformDestination(ctx mdformatter.SourceContext,
 		}
 	}
 
+	// Non md or image relative link, so needs link to be glued.
+	if !isMDFile(relDest) && !isImgFile(relDest) && r.glueLink != "" {
```
bwplotka, line 311: nit: We could try to check for `r.glueLink` first as it's the  most "important" one.

pkg/transform/transform.go
```diff
@@ -296,6 +307,30 @@ func (r *relLinkTransformer) TransformDestination(ctx mdformatter.SourceContext,
...
+		workingDir, err := os.Getwd()
+		if err != nil {
+			return nil, err
+		}
+		// Original path before transform needs to be figured out.
+		// Case where file is in same position after transform.
+		oldRelPath := filepath.Join(r.inputDir, curRelDir, relDest)
```
bwplotka, line 318: I think we do this twice or more.. wonder if we can have nice helper function for this

bwplotka (approved):
Nice work, small nits only otherwise LGTM 💪🏽

pkg/transform/testdata/mdox4.yaml
saswatamcode: Done!

pkg/transform/transform.go
saswatamcode: Done!

pkg/transform/transform.go
saswatamcode: I couldn't find a way to make a clean helper function for this as we do this multiple times but with some other argument each time. 😅

bwplotka (approved):
LGTM! Thanks 💪🏽

PR #78, opened 2021-08-22
Add retry for 0 StatusCode

saswatamcode:
This PR adds link retries for 0 status code. As seen in runs in Thanos (for example [here](https://github.com/thanos-io/thanos/runs/3368924522)), mdox errors out if there is status code 0. It isn't [defined](https://pkg.go.dev/net/http#pkg-constants) but generally occurs when the request is aborted for some reason or the host doesn't exist. 
Retrying once might help mitigate this and reduce the annoyance.

### DIFF
--- a/pkg/mdformatter/linktransformer/link.go
+++ b/pkg/mdformatter/linktransformer/link.go
@@ -284,7 +284,7 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 				break
 			}
 			v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(err, "%q rate limited even after retry; status code %v", response.Request.URL.String(), response.StatusCode)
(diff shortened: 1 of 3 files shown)


pkg/mdformatter/linktransformer/link.go
```diff
@@ -284,7 +284,7 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 				break
 			}
 			v.remoteLinks[response.Ctx.Get(originalURLKey)] = errors.Wrapf(err, "%q rate limited even after retry; status code %v", response.Request.URL.String(), response.StatusCode)
-		case http.StatusMovedPermanently, http.StatusTemporaryRedirect, http.StatusServiceUnavailable:
+		case http.StatusMovedPermanently, http.StatusTemporaryRedirect, http.StatusServiceUnavailable, 0:
```
bwplotka, line 287: // Let's add comment that 0 means error on call side.

bwplotka (approved):

bwplotka (approved):

Issue #22, opened 2021-05-12
Links with redirects result in flaky errors


saswatamcode:
Links like [this](https://github.com/thanos-io/thanos/blame/main/docs/getting-started.md#L118), result in flaky errors like below which are thrown only sometimes,
```
"https://blog.openshift.com/federated-prometheus-with-thanos-receive/" not accessible; status code 0: Get "https://www.openshift.com/blog/federated-prometheus-with-thanos-receive/": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```
Maybe having a custom timeout might help here?

douglascamata:
It would be great to tackle this one. Often it causes intermittent test build failures.

PR #86, opened 2022-07-11
Try to avoid intermittent errors when checking links

douglascamata:
Related to #22. 

- Stop relying on the `http.DefaultHTTPTransport`: it's effectively global state. It can be 
  modified by any other package. It has different default configuration depending on the 
  Go compiler version and we cannot trust that these defaults are good for us.

- Considering this is a tool often ran in resource constrained
  environments, like Github Actions, the HTTP client timeout was bumped
  to 30 seconds.

- Colly was configured with a depth of 1, to avoid crawling too much
  information from the websites

- The parallelism configuration of `colly` has been brought down from 100
  to 10. This setting means that at most 10 requests will be sent in
  parallel for the matching domains, instead of 100. This way we are 
  more friendly to the servers and hopefully they will be happier with us too, 
  returning more responses with 200 status code faster.

- Added a random delay of up to 1 second to create new requests to
  matching domains, again to be more friendly with the servers, etc,  etc.

## How did I test this?

I compiled a binary and ran the `make check-docs` target of the https://github.com/thanos-io/thanos project a couple of times without my changes and with them.

Without the changes it's easy to reproduce the failures indicated by #22. 

With the changes, I couldn't reproduce them.  

## How does it affect run time?

Locals runs were taking:

* Without changes: Failed local runs that failed were taking 2m48s. Successful local runs could be as fast as 1m31s.
* With changes
  * With 1s random delay: 3m20s.
  * Without random delay: 1m30s.

There are a lot of variables involved though, it isn't a reliable measurement. 

I think I could be able to get rid of the random delay to make this faster while still avoiding the intermittent failures, if that's desirable. WDYT, @bwplotka?

### DIFF
--- a/pkg/mdformatter/linktransformer/link.go
+++ b/pkg/mdformatter/linktransformer/link.go
@@ -222,6 +222,9 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 			KeepAlive: 30 * time.Second,
 		}).DialContext,
 	}
+	if config.HostMaxConns != nil {
+		transport.MaxIdleConnsPerHost = *config.HostMaxConns
+	}
 	v := &validator{
 		logger:         logger,
 		anchorDir:      anchorDir,
@@ -252,11 +255,17 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 	if config.Timeout != "" {
 		v.c.SetRequestTimeout(config.timeout)
 	}
-	if err := v.c.Limit(&colly.LimitRule{
-		DomainGlob:  "*",
-		Parallelism: 10,
-		RandomDelay: 1 * time.Second,
-	}); err != nil {
+
+	limitRule := &colly.LimitRule{
+		DomainGlob: "*",
+	}
+	if config.Parallelism != 0 {
+		limitRule.Parallelism = config.Parallelism
+	}
+	if config.RandomDelay != "" {
+		limitRule.RandomDelay = config.randomDelay
+	}
+	if err := v.c.Limit(limitRule); err != nil {
 		return nil, err
 	}
 	v.c.OnRequest(func(request *colly.Request) {
--- a/pkg/mdformatter/linktransformer/link.go
+++ b/pkg/mdformatter/linktransformer/link.go
@@ -208,10 +208,7 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 	}
 
 	linktransformerMetrics := newLinktransformerMetrics(reg)
-	collector := colly.NewCollector(colly.Async(), colly.StdlibContext(ctx), colly.MaxDepth(1))
-	collector.SetClient(&http.Client{
-		Timeout: 30 * time.Second,
-	})
+	collector := colly.NewCollector(colly.Async(), colly.StdlibContext(ctx))
 	transport := &http.Transport{
 		Proxy:                 http.ProxyFromEnvironment,
 		ForceAttemptHTTP2:     true,
--- a/pkg/mdformatter/linktransformer/link.go
+++ b/pkg/mdformatter/linktransformer/link.go
@@ -208,13 +208,30 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 	}
 
 	linktransformerMetrics := newLinktransformerMetrics(reg)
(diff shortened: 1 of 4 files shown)


pkg/mdformatter/linktransformer/link.go
```diff
@@ -241,7 +257,8 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 	}
 	if err := v.c.Limit(&colly.LimitRule{
 		DomainGlob:  "*",
-		Parallelism: 100,
+		Parallelism: 10,
```
bwplotka, line 260: > The parallelism configuration of colly has been brought down from 100
to 10. This setting means that at most 10 requests will be sent in
parallel for the matching domains, instead of 100. This way we are
more friendly to the servers and hopefully they will be happier with us too,
returning more responses with 200 status code faster.

Hope based development 🙈 

I think this makes sense generally, but I wonder if it makes sense to perhaps expose those options so we can tune it to every environment. Right now we are bit guessing with programmatic change that takes time to modify and release.

pkg/mdformatter/linktransformer/link.go
```diff
@@ -207,32 +207,48 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
 		}
 	}
 
+	linktransformerMetrics := newLinktransformerMetrics(reg)
+	collector := colly.NewCollector(colly.Async(), colly.StdlibContext(ctx), colly.MaxDepth(1))
```
saswatamcode, line 211: So setting `colly.MaxDepth` as 1 here has no effect for mdox. colly has the ability to scrape links from websites recursively, but mdox only uses the colly Collector.Visit [method](https://github.com/bwplotka/mdox/blob/main/pkg/mdformatter/linktransformer/validator.go#L55), which by [default always has a depth of 1](https://github.com/gocolly/colly/blob/master/colly.go#L493).

pkg/mdformatter/linktransformer/link.go
```diff
@@ -207,32 +207,48 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
...
+	collector.SetClient(&http.Client{
+		Timeout: 30 * time.Second,
```
saswatamcode, line 213: mdox already has a [parameter](https://github.com/bwplotka/mdox/blob/main/pkg/mdformatter/linktransformer/link.go#L240) in config which allows users to customize this, as different users will likely have different types of docs (how many links/type of links). 🙂 

We use this [parameter](https://github.com/prometheus-operator/prometheus-operator/blob/main/.mdox.validate.yaml#L2) in prometheus-operator too.

saswatamcode (commented):
Thanks a lot for taking this up! 💪🏻

pkg/mdformatter/linktransformer/link.go
```diff
@@ -207,32 +207,48 @@ func NewValidator(ctx context.Context, logger log.Logger, linksValidateConfig []
...
+	})
+	transport := &http.Transport{
```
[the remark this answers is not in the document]
douglascamata: Actually I don't think tuning `MaxIdleConns` to different values will be useful. But I found out `MaxConnsPerHost` (by default 2), which can be very useful. For example, Colly might be trying to send 10 requests to Github, but the transport will only hold 2 connections to Github.

pkg/mdformatter/linktransformer/link.go
douglascamata: Better, I am moving them to the config file, like the timeout.

pkg/mdformatter/linktransformer/link.go
douglascamata: Fixed in https://github.com/bwplotka/mdox/pull/86/commits/a02848d8d61d5cacc7455c3960df4a90387aa628

pkg/mdformatter/linktransformer/link.go
douglascamata: Added some parameters in the config file via https://github.com/bwplotka/mdox/pull/86/commits/23f1deca397085457bb4616ab93120a1175d6125.

douglascamata:
I made some parameters configurable via the validate config file and updated the example in the README to include them and the `timeout` parameter (it wasn't documented yet).

bwplotka (approved):
Amazing, thanks!

PR #88, opened 2022-07-12
Extend validator support to local links

### LINKED ISSUE #87: Add config option to skip local link checks
We currently can only ignore remote links based on regex, but there are use cases where users might also want to ignore local links.

saswatamcode:
This PR extends validator support for local links.

Fixes #87.

### DIFF
--- a/pkg/mdformatter/linktransformer/config.go
+++ b/pkg/mdformatter/linktransformer/config.go
@@ -20,9 +20,10 @@ import (
 type Config struct {
 	Version int
 
-	Validators  []ValidatorConfig `yaml:"validators"`
-	Timeout     string            `yaml:"timeout"`
-	Parallelism int               `yaml:"parallelism"`
+	ExplicitLocalValidators bool              `yaml:"explicitLocalValidators"`
+	Validators              []ValidatorConfig `yaml:"validators"`
+	Timeout                 string            `yaml:"timeout"`
+	Parallelism             int               `yaml:"parallelism"`
 	// HostMaxConns has to be a pointer because a zero value means no limits
 	// and we have to tell apart 0 from not-present configurations.
 	HostMaxConns *int   `yaml:"host_max_conns"`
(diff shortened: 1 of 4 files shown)


pkg/mdformatter/linktransformer/config.go
```diff
@@ -52,12 +55,22 @@ type IgnoreValidator struct {
 	_regex *regexp.Regexp
 }
 
+type LocalValidator struct {
+	_regex *regexp.Regexp
+}
+
+type IgnoreLocalValidator struct {
+	_regex *regexp.Regexp
```
bwplotka, line 63: Let's use exactly the same validator routines for local links, so we don't need to add any extra config bit.

Probably to ensure compatibility with current use cases, we might want to add special flag for it `--enable-explicit-local-validators`

bwplotka (commented):
Thanks! Some idea first to check (:

pkg/mdformatter/linktransformer/config.go
saswatamcode: Keeping it as a YAML config option instead, as we mention validators via that too. 🙂

pkg/mdformatter/linktransformer/validator.go
```diff
@@ -32,8 +32,31 @@ func (v GitHubPullsIssuesValidator) IsValid(k futureKey, r *validator) (bool, er
 	return false, nil
 }
 
-// RoundTripValidator.IsValid returns true if url is checked by colly.
+// RoundTripValidator.IsValid returns true if url is checked by colly or it is a valid local link.
 func (v RoundTripValidator) IsValid(k futureKey, r *validator) (bool, error) {
+	matches := remoteLinkPrefixRe.FindAllStringIndex(k.dest, 1)
+	if matches == nil && r.validateConfig.ExplicitLocalValidators {
```
bwplotka, line 38: Does it make sense to pack it in separate function? Looks like 1:1 to repetition.

bwplotka (approved):
LGTM, just one nit, not a blocker - feel free to merge.

bwplotka (approved):

PR #91, opened 2022-07-13
go.{mod,sum}: bump Go dependencies

simonpasquier:
This change bumps the following Go modules to get rid of some
vulnerability reports:
* https://github.com/advisories/GHSA-x95h-979x-cf3j
* https://github.com/advisories/GHSA-8j34-9876-pvfq

Though mdox is probably not affected, it will avoid dependabot reporting
security issues in mdox and all projects depending on it.

Signed-off-by: Simon Pasquier <spasquie@redhat.com>


bwplotka (approved):
Thanks 💪🏽

PR #93, opened 2022-07-15
Parse cache settings manually for friendly errors

douglascamata:
Just like it's done for all the other configuration options that
are based off of time durations.

Signed-off-by: Douglas Camata <159076+douglascamata@users.noreply.github.com


saswatamcode (approved):
Thanks for making it consistent!

PR #94, opened 2022-07-29
Isolate and refactor caching logic

douglascamata:
Signed-off-by: Douglas Camata <159076+douglascamata@users.noreply.github.com>

The cache logic has been moved to its own package.

The refactor improves on: 

- Using a typed string for the cache type, to provide some additional safety. 
- Makes it harder to access the underlying string fields of configuration options that are time durations to, to help avoiding mistakes.

## Breaking change 

* This PR changes the SQLite cache type configuration option from `SQLite` to `sqlite`. Helps avoid issues with upper/lower case mistakes.


pkg/cache/config.go
```diff
@@ -0,0 +1,101 @@
...
+	if err := c.load(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// load validates the cache configuration from the parser and copy it
+// into the configuration.
+func (c *Config) load() error {
+	switch c.cacheParser.Type {
```
bwplotka, line 62: Can we lower case the Type here, so we can make this change but also without breaking change?

pkg/cache/config.go
```diff
@@ -0,0 +1,101 @@
...
+
+// NewConfig is the constructor for Config.
+func NewConfig() Config {
+	return Config{
+		cacheParser: newConfigParser(),
+	}
+}
+
+// Present tell whether a cache configuration is present.
+func (c *Config) Present() bool {
```
bwplotka, line 43: "present" 


(: 

What about: 
```suggestion
func (c *Config) IsSet() bool {
```

pkg/cache/config.go
```diff
@@ -0,0 +1,101 @@
...
+	return c.Type != cacheTypeNone
```
bwplotka, line 44: Also 
```suggestion
	return c.Type != cacheTypeNone && c.Type != cacheTypeEmpty
```

bwplotka (approved):
Thanks! Some nits, otherwise LGTM

pkg/cache/config.go
douglascamata: @bwplotka I can un-export the `Type`, but I need a getter to see the actual type. What's the intent of this suggested change?

douglascamata: Ah, nevermind. I was confused.

douglascamata:
@bwplotka feedback taken care of. 👍 🎉

saswatamcode (approved):
Thanks!

PR #97, opened 2022-10-07
Drop ioutil

abhinav:
ioutil was deprecated in Go 1.16.
Drop all usage of ioutil in favor of non-ioutil variants.


saswatamcode (approved):
Thanks! LGTM!

PR #96, opened 2022-10-07
Use Go 1.19

abhinav:
This raises the `go` directive in the go.mod file to Go 1.19,
and runs `go mod tidy`,
which prunes a fair bit of redundant information from the go.sum.

Additionally, this bumps the version of Go used in both GitHub
workflows.

PR #98, opened 2022-10-11
Add --no-code-fmt option

abhinav:
Kunde21/markdownfmt attempts to reformat all snippets using the "go"
language type.

So it will rewrite anything like the following:

    ```go
    package main

    func foo() {
      // intentionally using two spaces
    }
    ```

This change adds a --no-code-fmt command line option that opts out of this
behavior in markdownfmt by replacing the list of code formatters.

The option is named no-code-fmt instead of no-gofmt to account
for other code formatters that mdox may introduce in the future.

This change requires Kunde21/markdownfmt#38.
This change also depends on #96.


saswatamcode (commented):
Not sure I understand the motivations behind this change. 

Why not reformat Go code with gofmt? It's enabled by default on every IDE and it is by design opinionated and non-configurable.

Based on PRs you linked, https://github.com/uber-go/fx/pull/961, you mention that the gofmt formatting (via mdox) is inconsistent. Maybe addressing that issue, instead of disabling it altogether would be a better approach here? Wdyt? 🙂

abhinav:
Rebased on top of #96 and dropped the temporary pin-to-fork.

abhinav:
Just to elaborate on https://github.com/bwplotka/mdox/pull/98#issuecomment-1277871894 with some example scenarios:

- Incremental step-by-step instructions don't always have valid code. For example, imagine something like,

    1. Start a new main function.

        ```go
        func main() {
        ```

    2. Initialize a new application.

        ```go
        func main() {
            app := fx.New(
            )
        ```
    
    3. ...and so on
    
    In these cases, we won't have valid Go code until we get to the end,
    and gofmt will not reformat anything until that point.

- In a style-guide context, you want to demonstrate good vs bad examples
  where having no reformatting is desirable.

(FWIW, all our code samples are coming from external files that are properly gofmt-ed.)

go.mod
```diff
@@ -1,10 +1,9 @@
 module github.com/bwplotka/mdox
 
-go 1.15
+go 1.19
```
saswatamcode, line 3: I think you raised PR for this separately #96?

pkg/mdformatter/mdformatter_test.go
```diff
@@ -147,3 +147,18 @@ func TestCheck_SoftWraps(t *testing.T) {
 	testutil.Ok(t, err)
 	testutil.Equals(t, string(exp), diff.String())
 }
+
+func TestFormat_NoGofmt(t *testing.T) {
```
saswatamcode, line 151: Thanks for the separate test case. I would also check what happens to this go code sample when --no-gofmt is disabled, just to ensure we aren't changing default behavior

saswatamcode (commented):
> In a style-guide context, you want to demonstrate good vs bad examples
where having no reformatting is desirable.

This use case makes sense to some extent, but would a style guide redefine what is already covered by gofmt?

Anyways, it can still be a useful debugging option, so I do not want to block it! Let's wait for @bwplotka's opinions too. 🙂 

Also some small suggestions!

bwplotka:
👍🏽  I think it definitely makes sense to be able to disable gofmt - solid use cases. Perhaps doing this with extra directive on code snippet would do? Essentially opt-out for formatting per snippet?

go.mod
abhinav: Yeah, this PR is on top of that one. If #96 gets merged, this won't show those commits.
(Without stacking them, there are conflicts because of the dependency updates.)

abhinav:
> 👍🏽 I think it definitely makes sense to be able to disable gofmt - solid use cases. Perhaps doing this with extra directive on code snippet would do? Essentially opt-out for formatting per snippet?

Thanks @bwplotka, I can add that to this PR too.
Are you still okay with the `--no-gofmt` top-level flag?
For our documentation, we'll otherwise end up adding the per-snippet opt-out on all snippets.

abhinav:
I can call this `--no-fmt` and the per-snippet-opt-out `mdox-fmt="false"`.
This way, if mdox adds reformatting for other languages, the same opt-out works.

saswatamcode:
I think flag + directive makes sense here! The `no-fmt` flag would basically switch default to `mdox-fmt=true`.

abhinav:
On second thought, per-snippet opt-out is possible,
but it either requires new APIs in markdownfmt to have a hook before it runs a reformat,
or we change it so that mdox always opts out of reformatting from markdownfmt,
and does its own reformatting if necessary.

For now, I've updated this PR with additional tests as suggested,
and renamed the flag to `--no-code-fmt` to opt out of all code reformatting.
This will account for new code reformatting logic being added in the future.

abhinav:
@bwplotka @saswatamcode 
Any appetite for file-level opt-out with `--no-code-fmt`?
I don't know if I have time to implement per-snippet opt-out in markdownfmt at this time.

saswatamcode:
Ack! WIll try to TAL this week, and see what would be nice to implement here! Thanks for the awesome work! 🙂

abhinav:
Superseded by #99

PR #99, opened 2022-12-12
Upgrade markdownfmt, add opt-out for gofmt

abhinav:
markdownfmt v3 was recently released,
and among the changes is how Go source blocks are handled.
They're no longer automatically run through gofmt.
Instead, users may opt-into this functionality with the following option:

    markdown.WithCodeFormatters(markdown.GoCodeFormatter)

Since mdox prefers to have Go blocks auto-formatted,
this PR enables that by default,
and adds an opt-out flag (`--no-code-fmt`)
for users who don't want that functionality.

Supersedes #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Flags:
                                  instead it will fail if files needs formatting
       --soft-wraps               If true, fmt will preserve soft line breaks for
                                  given files
+      --no-code-fmt              If true, don't reformat code snippets
       --code.disable-directives  If false, fmt will parse custom fenced
                                  code directives prefixed with 'mdox-gen' to
                                  autogenerate code snippets. For example:
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/bwplotka/mdox
 go 1.19
 
 require (
-	github.com/Kunde21/markdownfmt/v2 v2.1.1-0.20221212171616-95b15aa5ffe4
+	github.com/Kunde21/markdownfmt/v3 v3.0.0
 	github.com/charmbracelet/glamour v0.5.0
 	github.com/efficientgo/tools/core v0.0.0-20210609125236-d73259166f20
 	github.com/efficientgo/tools/extkingpin v0.0.0-20210609125236-d73259166f20
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/bwplotka/mdox
 go 1.19
 
 require (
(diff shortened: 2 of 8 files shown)


bwplotka:
Nice! I remember we wanted or maybe it's there - the way to change gofmt formatting per code source basis too, do you remember @saswatamcode? 

Otherwise generally looks good, thanks!

abhinav:
Hey, @bwplotka, thanks for looking.

> I remember we wanted or maybe it's there - the way to change gofmt formatting per code source basis too

Yeah, we discussed the possibility of that in #98. A good design for it in markdownfmt wasn't obvious so I deferred it at the time. After thinking about it more, I've got a couple points I'd like to make:

1. We should reconsider whether per-block opt-outs are actually a feature users will use. In my opinion, for the people who want this, it's an all-or-nothing situation: they either want all blocks reformatted or none of them.
    1. Just to add another use case, the reason the folks at [uber-go/fx](https://github.com/uber-go/fx) wanted full opt-out was this: code snippets in their [documentation](https://uber-go.github.io/fx/) are extracted from tested files and run through a [filter](https://github.com/uber-go/fx/blob/3a86277ab0e5d84cfee67a53b4a0f15c824a80a4/docs/bin/region) that transforms tabs to 2 spaces so as to reduce horizontal space lost to long lines. Having mdox reformat them means that some code samples (the ones that are completely valid code) end up reformatted back to tabs, and everything else remains using 2 spaces.
2. If we do want per-block opt-outs, it's my opinion that from the point of view of layer boundaries, markdownfmt should not try to interpret code block info strings. markdownfmt should try to be just a Markdown formatter/renderer. Selective formatting logic should be implemented inside mdox. One way to go about this is: mdox never enables markdownfmt's code formatting, and instead transforms the AST on its end, reformatting code blocks as needed, before calling markdownfmt to render the AST. That said, I don't think I have the time to implement this functionality in mdox right now.

Regardless of the above, I'd like to ask that you please consider this PR (with global opt-out support) separate from the discussion of per-block opt-outs. Global opt-outs do not obviate the possibility of per-block opt-in/opt-out as a feature. Even if we fully agreed on per-block opt-outs support, it would make sense for it to be a separate PR, and not part of this one.

bwplotka:
Looks good, should we proceed @saswatamcode ?

saswatamcode (approved):
Sorry for delay, I missed notifs for this.

Thanks for great work and explanation above! Seems that we should proceed with this then! 🙂

PR #103, opened 2023-01-24
mdox-exec: Also add newline when non-zero exit code

olivierlemasle:
A new line (`\n`) is added at the end of `mdox-exec` output, but only when the exit code is 0 (not when using a non-zero `mdox-expect-exit-code`):

https://github.com/bwplotka/mdox/blob/53a5104b2ef519f3e35ebb0605c86121c503bc38/pkg/mdformatter/mdgen/mdgen.go#L92-L95

The new line should always be added (if not present).

Tests are updated to use `echo -n` (prints no `\n`).

### DIFF
--- a/pkg/mdformatter/mdgen/mdgen.go
+++ b/pkg/mdformatter/mdgen/mdgen.go
@@ -83,10 +83,13 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 		cmd.Stdout = &b
 		if err := cmd.Run(); err != nil {
 			expectedCode, _ := strconv.Atoi(infoStringAttr[infoStringKeyExitCode])
-			if exitErr, ok := err.(*exec.ExitError); ok && exitErr.ExitCode() == expectedCode {
-				return b.Bytes(), nil
+			if exitErr, ok := err.(*exec.ExitError); ok {
+				if exitErr.ExitCode() != expectedCode {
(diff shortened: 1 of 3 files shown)


pkg/mdformatter/mdgen/mdgen.go
```diff
@@ -83,10 +83,9 @@ func (t *genCodeBlockTransformer) TransformCodeBlock(ctx mdformatter.SourceConte
 		cmd.Stdout = &b
 		if err := cmd.Run(); err != nil {
 			expectedCode, _ := strconv.Atoi(infoStringAttr[infoStringKeyExitCode])
-			if exitErr, ok := err.(*exec.ExitError); ok && exitErr.ExitCode() == expectedCode {
-				return b.Bytes(), nil
+			if exitErr, ok := err.(*exec.ExitError); !ok || exitErr.ExitCode() != expectedCode {
```
saswatamcode, line 86: Maybe we should return a separate error if `!ok` and `exitErr.ExitCode() != expectedCode`?

saswatamcode (approved):
Thanks for fixing!

One small suggestion. 🙂

pkg/mdformatter/mdgen/mdgen.go
olivierlemasle: Thanks, I've updated the PR to differentiate these errors.

saswatamcode (approved):
Thanks!

PR #108, opened 2023-05-07
Update build

SuperQ:
* Enable dependabot.
* Update Go to 1.20.
* Update bingo.
* Update golangci-lint.
* Remove misspell (handled by golangci-lint).


saswatamcode (approved):
Thanks a lot! Generally LGTM! 

I think the "check 404 link" test fails now, due to the new go version printing DNS resolver address with the error. Maybe we can convert https://github.com/bwplotka/mdox/blob/main/pkg/mdformatter/linktransformer/link_test.go#L311 to use `testutil.Assert` with `strings.Contains` instead, as that might be simplest way to fix? 🙂

SuperQ:
@saswatamcode Done, updated test.

pkg/mdformatter/linktransformer/link_test.go
```diff
@@ -308,9 +304,7 @@ func TestValidator_TransformDestination(t *testing.T) {
 			MustNewValidator(logger, []byte(""), anchorDir, nil),
 		))
 		testutil.NotOk(t, err)
-		testutil.Equals(t, fmt.Sprintf("%v: 2 errors: "+
-			"%v:1: \"https://docs.gfoogle.com/drawings/d/e/2PACX-1vTBFK_cGMbxFpYcv/pub?w=960&h=720\" not accessible even after retry; status code 0: Get \"https://docs.gfoogle.com/drawings/d/e/2PACX-1vTBFK_cGMbxFpYcv/pub?w=960&h=720\": dial tcp: lookup docs.gfoogle.com: no such host; "+
-			"%v:1: \"https://bwplotka.dev/does-not-exists\" not accessible; status code 404: Not Found", tmpDir+filePath, relDirPath+filePath, relDirPath+filePath), err.Error())
+		testutil.Assert(t, strings.Contains(err.Error(), "status code 404: Not Found"))
```
saswatamcode, line 307: Thanks! This test is meant to check if we are including filename, line number and the faulty link in the error, so maybe we can do something like this? 🙂  
```suggestion
		testutil.Assert(t, strings.Contains(err.Error(), fmt.Sprintf("%v:1: \"https://docs.gfoogle.com/drawings/d/e/2PACX-1vTBFK_cGMbxFpYcv/pub?w=960&h=720\" not accessible even after retry; status code 0",relDirPath+filePath)))
		testutil.Assert(t, strings.Contains(err.Error(), fmt.Sprintf("%v:1: \"https://bwplotka.dev/does-not-exists\" not accessible; status code 404: Not Found", relDirPath+filePath)))
		
```

saswatamcode: I'll commit this and merge!

saswatamcode:
Ah missed the path!

pkg/mdformatter/linktransformer/link_test.go
SuperQ: Ahh, thanks, I thought the test was just for the 404 as the name suggested.

SuperQ:
One thing that can also be done now is switch to the Go 1.20 multi-error handling.

PR #115, opened 2023-05-08
Update go-kit logging

SuperQ:
The library `github.com/go-kit/kit/log` has been replaced by `github.com/go-kit/log`.
* Use inline errorcheck excludes.


saswatamcode (approved):
Thanks for updating this!

PR #116, opened 2023-05-08
Update kingpin library

SuperQ:
The kingpin library moved to a new path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,56 +37,58 @@ usage: mdox fmt [<flags>] <files>...
 Formats in-place given markdown files uniformly following GFM (Github Flavored
 Markdown: https://github.github.com/gfm/). Example: mdox fmt *.md
 
+
 Flags:
-  -h, --help                     Show context-sensitive help (also try
-                                 --help-long and --help-man).
-      --version                  Show application version.
-      --log.level=info           Log filtering level.
-      --log.format=clilog        Log format to use.
+  -h, --[no-]help              Show context-sensitive help (also try --help-long
+                               and --help-man).
+      --[no-]version           Show application version.
+      --log.level=info         Log filtering level.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -37,56 +37,58 @@ usage: mdox fmt [<flags>] <files>...
...
-      --anchor-dir=ANCHOR-DIR    Anchor directory for all transformers. PWD is
-                                 used if flag is not specified.
+                               Path to directory where metrics are saved in
+                               OpenMetrics format; If empty, no metrics will be
+                               saved.
+      --[no-]check             If true, fmt will not modify the given files,
+                               instead it will fail if files needs formatting
+      --[no-]soft-wraps        If true, fmt will preserve soft line breaks for
+                               given files
+      --[no-]no-code-fmt       If true, don't reformat code snippets
```
saswatamcode, line 59: This ends up being somewhat confusing now. Seems like new kingpin version adds `[no-]` to all bool flags as a way of passing false values, which breaks some naming like here.

I'm not entirely sure what we can do here, other than changing some of the flag names, and true/false conventions so that the `--no-*` flags make sense here. 
I'll put some thought into it and try to merge and iterate on this!

Might also be good to do this in other places like Thanos. cc: @bwplotka

SuperQ: This has always been the way the kingpin library worked. The change is that it's now documented in the `--help`.

SuperQ: Want me to make this bool change here? Or a separate PR?

saswatamcode (approved):
Thanks! LGTM! 🙂

README.md
bwplotka: Hm, I think it was always the case 🤔

bwplotka: I think only help is buggy or our flags wasn't considering that

PR #158, opened 2024-04-01
fix typos

ccoVeille:
I started looking at your code and I noticed a few typos.

I was about to open a simple PR, then I found acronyms incorrectly used, so I fixed them.


pkg/mdformatter/linktransformer/link.go
```diff
@@ -551,7 +551,7 @@ func (l localLinksCache) addRelLinks(localLink string) error {
 func toHeaderID(header []byte) string {
 	var id []byte
 	// Remove punctuation from header except '-' or '#'.
-	// '\p{L}\p{N}\p{M}' is the unicode equivalent of '\w', https://www.regular-expressions.info/unicode.html.
+	// '\p{L}\p{N}\p{M}' is the Unicode equivalent of '\w', https://www.regular-expressions.info/unicode.html.
```
bwplotka, line 554:
```suggestion
	// '\p{L}\p{N}\p{M}' is the unicode equivalent of '\w', https://www.regular-expressions.info/unicode.html.
```

This one is correct honeslty

bwplotka (approved):
Thanks!

pkg/mdformatter/linktransformer/link.go
bwplotka: but not a big deal

